PR #53, opened 2019-06-30
Many documentation corrections to the User Guide

enbrown:
Here are many, many fixes to the user guide documentation.

The main ones in no particular order:

* Changed tables of equations to `\begin{align*} ... \end{align*}` environments to align related equations. Also changed the formatting/markup for cases enviornments to be more consistent with spacing and font styling.

* Removed extra blank lines above LaTeX equations. An extra blank line before an equation instructs LaTeX that the equation is the beginning of a new paragraph and thus includes too much space above the equation in the PDF otuput.

* Rewrote some of the equations with the nonstandard 1:N form as $\{1,\dotsc,N\}$ instead.

* Used the \dotsc rather than \ldots when the dots are to separate comma lists. Some places used \dotsb to separate binary operators.

* Changed all uses of \mbox to either \operatorname (for things like I() where I() is a function similar to how exp() is a  function), \mathtt when the name is a Stan command, and \mathsf where the name is a distribution. For the last, if the whole document was typed in a serif-font, then \mathrm might be better; but since the entire document is typeset in a sans-serif font, \mathsf is likely better. Some of these choices were arbitrary but I attempted to use the surrounding context to tell whether it was denoting a mathematical function (\operatorname), a Stan function (\mathtt), or a statistical distribution (\mathsf).

* Cleaned up some LaTeX that had found its way into the verbatim (`Stan code here`) sections.

* Fixed some whitespace issues with Stan verbatim blocks. If there were extra whitespace at the beginning of every line in a verbatim block, this was removed because it would increase the indentation arbitrarily in the PDF output.

* Removed some hold-over id:... lines. I'm not sure the purpose of these so I converted most to HTML anchors (<a name="id:stuff..."></a>) in case they are needed later.

* Replaced p(y|x) with p(y \mid x) in LaTeX equations to correctly space the pipe in the middle

* Removed a few mispalced ~ symbols (which are non-breaking spaces in LaTeX but not in Markdown)

* Replaced a few * symbols with \times in LaTeX equations

* Cleaned up some of the formatting of equations to make the parentheses nicer. This used `\left(...\right)` when appropriate and sometimes `\big(...\big)` and similar commands.

* Fixed a few typos to the missing-data.Rmd file.

* Fixed some equation typos in truncation-censoring.Rmd. One of the mentions of normal_l
ccdf in an equation was used as `normal_lccdf` in an equation environment which ended
up looking very strange. This was changed to \mathtt{normal\_lccdf} to be properly for
matted. It was also placed in an align* environment and broken into two lines to bette
r flow.

* Fixed some equation formatting for measurement-error.Rmd.

Made many fixes to latent-discrete.Rmd:

* Recreated a table in HTML format because Markdown in pandoc is incapable of handling cells that span multiple columns. Care was taken to mimic the LaTeX formatting and enable the embedded symbols to be processed by MathJax.

* Converted embedded PDF images to PNG format and then updated the knitr::include_graphics() call to use the PDF version when it exists and can be used in the output (such as LaTeX/PDF output.) The command `convert -density 300 foobar.pdf foobar.png` was used for the conversion.

* Split many equations across multiple lines because they were too long to be printed on one line in the LaTeX output. This used the `align*` LaTeX environment extensively.

* Changed the use of many non-name distributions (such as uniform) to be lower-cased. The style-guide (somewhere) states that named distributions (Poisson for example) are to be in title-case whereas the non-name distributions (including normal and uniform) should be in lowercase.

Made some changes to sparse-ragged.Rmd:

* Created an HTML version of the first LaTeX minipage figure.

* Converted the second minipage figure into a align* enviornment and adjusted the wording so that the one set of markup will work for both HTML and LaTeX.

* Used bmatrix environments rather than coding brackets in the equations to better handle spacing.

Updated clustering.Rmd

* Fixed a typo with --- (there aren't supposed to be spaces before/after the em-dash)

* Normalized title-case vs lowercase for distribution names such as categorical. Because the LKJCorr density is defined by 3 named individuals, I used LKJCorr instead of LkjCorr or lkjcorr. I'm not sure that this is right or standard, but just guessing.

* Created an HTML table replacement for the LaTeX table of coding ragged arrays.

Cleaned up gaussian-processes.Rmd:

* Changed the typography of the cholesky_decompose function in equations to use \mathtt{} to match how other Stan functions are represented in equations.

Fixed typos in hyperspherical-models.Rmd:

* In one place $pi$ was used rather than $\pi$ as expected. In another `real <lower=-pi(),upper=pi>` was used whereas `real <lower=-pi(),upper=pi()>` is likely correct.

* Fixed some minor (likely not noticable) spacing around colons (:).

* Added closed curly-braces `}` to some of the blocks in the last Stan example.

Made a few minor changes to algebraic-equations.Rmd:

* A few places \textit{} was used to attempt to italicize text. This works for the LaTeX output but makes the text disappear for the HTML output. This was changed to the *emphasis here* Markdown syntax.

* Clarified in the Control Parameters what the 1e-10, 1e-6, and 1e3 defaults were.

Made corrections to odes.Rmd:

* Included PNG version of sho-ode-trajectory.pdf for the HTML user guide.

* Cleaned up some of the typography of the LaTeX equations (such as using a bmatrix environment for matrices, \qquad for horizontal spacing rather than a bunch of spaces, and a few scattered \, spaces to better space matrix multiplcations.)

Fixed some typos in floating-point.Rmd:

* There were a few exponents and a few subscripts that were displaying incorrectly because 10^52 needs to be written in TeX as $10^{52}$ otherwise only the 5 will show up as a superscript (similarly with subscripts such as $\log_{10}$ for log-based-10).

* The decimal representation of 36.29e-3 was incorrect.

* Changed some newlines so that equations such as `0 + 1e-20 == 1e-20` would show up on one line rather than having a newline break it across two lines.

* Fixed a missing back-slash in $L^{\top}$ which was written as $L^{top}$.

* Use a LaTeX equation to show catastrophic cancellation rather than a verbatim environment.

* Changed log1p and log_sum_exp to use a typewriter font (\mathtt) to match other Stan functions.

* Aligned comments in Stan blocks in multi-indexing.Rmd

Fixed a few minor issues with user-functions.Rmd:

* Used \mathrm{I} instead of \mbox{I} and \text{...} in the \begin{cases}...\end{cases} environment.

* Left-aligned the code in the constrainted-types table.

* Removed erroneous \, characters in the Array Types for Function Delarations section.

Fixed a few issues with custom-probability.Rmd:

* The binormal_cdf was using an \mbox and thus showing up incorrectly.

Fixed some issues with problematic-posteriors.Rmd:

* Used `\mathrm{I}` in places (the more consistent \mathsf{I} wouldn't have the serifs which makes the I difficult to differentiate from an L or 1).

* Added subtle spacing for `1\,000\,000\,000`.

* Fixed a misplaced prime in `\exp(\alpha_{k'})`

Fixed some issues with reparameterization.Rmd:

* There were a few places where $\mathrm raw$ was used rather than $\mathrm{raw}$

* There was a missing closing parentheses

Made some fixes to efficiency-tuning.Rmd:

* Removed some extra space in $Student-t$ formatting using the negative space \! LaTeX markup.

* Used a pmatrix environment for the matrix used in describing the Cholesky factor of the scale matrix

* Tried to trim down the space taken by a fraction in the text by using a slanting fraction

* Cleaned up the equations on standardized predictors and outcomes

Made a few minor tweaks to map-reduce.Rmd:

* Line-wrapped overly long Stan block lines.

#### Submission Checklist

- [ ] Builds locally --- it does for me but I don't have the Lucida font to check

#### Copyright and Licensing

Please list the copyright holder for the work you are submitting (this will be you or your assignee, such as a university or company): Eric N. Brown


bob-carpenter:
Thanks so much!  This is really really great.  I'm rather ashamed at what a mess it was given how much you needed to fix.

I only have a few requested changes, which I've marked with boxes---the rest are comments.   I'm going to state these at a high level, then if you want to go ahead and make some of the changes, I can do a fine-grained review of the final submission.

The biggest problem we're going to run into is that you've adopted new conventions for so many things that the reference manual won't match in style any more, so I guess we'll have to fix that to match your changes.  Do you have a list of them anywhere other than here?  We don't have a style guide for the docs.

> since the entire document is typeset in a sans-serif font

[ ] Optional, because I don't even know if it's possible and it's beyond the scope of your PR.  
Serif should be the default in html.  The pdfs are serif.  I realize that I set my own output to serif form and my browser (Safari) remembers it between sessions.  I'd like to make serif the default if possible as the sans serif is both ugly and hard to read.

> Changed tables of equations to \begin{align*} ... \end{align*}

Is that going to reduce space around =?  I find the default's too tight, which is why I liked using arrays (I could've also used eqnarray, but I found it dicey in the past versus using their $$ thing).  This isn't a deal breaker, though.

> Rewrote some of the equations with the nonstandard 1:N form as ${1,\dotsc,N}$ instead.

I wrote them as 1:N on purpose.  Not a big deal, but this kind of style thing will just keep getting changed as everyone looks at it.  Why not n = 1 below and N above?  That's the only thing that's standard in intro calc books.

> \dotsc rather than \ldots when the dots are to separate comma lists. Some places used \dotsb to separate binary operators.

<b>?</b> Never heard of these.  What's the effect?

> Changed all uses of \mbox to either \operatorname (for things like I() where I() is a function similar to how exp() is a function), \mathtt when the name is a Stan command, and \mathsf where the name is a distribution. 

I agree that where I used `\mbox` should be replaced.  The operator names are nice, but the other replacements should be the text forms to get the inter-letter spacing right.  Math messes it up.  

[ ] So that should be `texttt` and `textsf` and `textrm`.

> Removed some hold-over id:... lines. I'm not sure the purpose of these so I converted most to HTML anchors () in case they are needed later.

Thanks.  It was a real struggle to get the cross-references working.  I'll have to see how you did it.

> Replaced p(y|x) with p(y \mid x) in LaTeX equations to correctly space the pipe in the middle

This is a contentious typesetting issue, but I agree.  I've been switching to use `\mid` almost everywhere in new things I write.

> Recreated a table in HTML format because Markdown in pandoc is incapable of handling cells that span multiple columns. Care was taken to mimic the LaTeX formatting a

Wow.  Thanks.

> Converted embedded PDF images to PNG format and then updated the knitr::include_graphics() 

Cool.  And thanks!  I think 300dpi is right for rendering.  Do you have any kind of automatic png converter from LaTeX?  I know how to do it using a drag-and-drop interface, but that's a huge pain.  

What we really need to do is generate those images using markdown!

> Split many equations across multiple lines because they were too long to be printed on one line 

Thanks.  As you can imagine, this version targeted the html and neglected the pdf. 

> Changed the use of many non-name distributions (such as uniform) to be lower-cased. The style-guide (somewhere) states that named distributions (Poisson for example) are to be in title-case whereas the non-name distributions (including normal and uniform) should be in lowercase.

This has been totally inconsistent everywhere.  I'd be just as happy lowercasing "Poisson" as then it'd match how things are rendered in Stan.

> there aren't supposed to be spaces before/after the em-dash

I agree we should stick to American typesetting and spelling just to be consistent (they add spaces in the UK, which is where I learned to LaTeX and I'm still confused on this and a number of other points [like what to call the room with a toilet] 30 years later).

> Changed the typography of the cholesky_decompose function in equations to use \mathtt{}

I want to use `textrm` for math symbols.  `texttt` should be reserved for computer program text.  So we write `\textrm{square}` for the mathematical function and `\texttt{square}` when talking about the Stan function.  Then it's tricky as to which symbol to use on the left-hand side of a definition, because mathematical functions don't take NaN inputs or +/- infinity.  So I don't know what to do here, so I'm not too bothered by changes as long as they're consistent.

enbrown:
> [ ] Optional, because I don't even know if it's possible and it's beyond the scope of your PR.
> Serif should be the default in html. The pdfs are serif. I realize that I set my own output to serif form and my browser (Safari) remembers it between sessions. I'd like to make serif the default if possible as the sans serif is both ugly and hard to read.

Sorry, I didn't have the Lucida font that `header.tex` referenced and when I ran it using the build script, my LaTeX pulled in a sans-serif font so I thought that was the one to use. I should have checked it against the one on mc-stan.org (which is clearly serif). So, some of the \mathsf{} markup might be better as \mathrm{}.

>
> > Changed tables of equations to \begin{align*} ... \end{align*}
> 
> Is that going to reduce space around =? I find the default's too tight, which is why I liked using arrays (I could've also used eqnarray, but I found it dicey in the past versus using their $$ thing). This isn't a deal breaker, though.

This shouldn't reduce the spacing but rather make it the same as most LaTeX typesetting. Some of the alternatives have been found to be inconsistent so I think that the LaTeX community generally recommends `align` now.

> 
> > Rewrote some of the equations with the nonstandard 1:N form as ${1,\dotsc,N}$ instead.
> 
> I wrote them as 1:N on purpose. Not a big deal, but this kind of style thing will just keep getting changed as everyone looks at it. Why not n = 1 below and N above? That's the only thing that's standard in intro calc books.
> 

Again, sorry, I thought that this was more R nomenclature rather than standard math now.

> > \dotsc rather than \ldots when the dots are to separate comma lists. Some places used \dotsb to separate binary operators.
> 
> ? Never heard of these. What's the effect?
> 

Better spacing (https://en.wikibooks.org/wiki/LaTeX/Mathematics#Dots). I think that somewhere along the way different journals used different spacing between commas vs plus-signs and so these came about. These are in the CTAN `amsmath` package defines them but I've never been able to understand TeX code well enough to parse the difference.

> > Changed all uses of \mbox to either \operatorname (for things like I() where I() is a function similar to how exp() is a function), \mathtt when the name is a Stan command, and \mathsf where the name is a distribution.
> 
> I agree that where I used `\mbox` should be replaced. The operator names are nice, but the other replacements should be the text forms to get the inter-letter spacing right. Math messes it up.
> 
> [ ] So that should be `texttt` and `textsf` and `textrm`.
> 

Makes sense.

> > Removed some hold-over id:... lines. I'm not sure the purpose of these so I converted most to HTML anchors () in case they are needed later.
> 
> Thanks. It was a real struggle to get the cross-references working. I'll have to see how you did it.
> 

Basically I didn't try to get it working (yet). I tried not to remove any anchors but didn't try to get it to actually work.

> > Replaced p(y|x) with p(y \mid x) in LaTeX equations to correctly space the pipe in the middle
> 
> This is a contentious typesetting issue, but I agree. I've been switching to use `\mid` almost everywhere in new things I write.
> 

There were a couple places where something like `p\left( y \middle| x \right)` might be better, but for most places, I took your lead and changed them to `\mid` (which I didn't know worked until I came across yours.

> > Recreated a table in HTML format because Markdown in pandoc is incapable of handling cells that span multiple columns. Care was taken to mimic the LaTeX formatting a
> 
> Wow. Thanks.

I first used HTML tables for laying out webpages, so this took me back. I did try to use the latest syntax so they'll look nice in (and maybe print nice from) browsers. But copying the LaTeX table formatting (which I prefer) was just difficult given CSS rules overriding so the formatting isn't the same.

> > Converted embedded PDF images to PNG format and then updated the knitr::include_graphics()
> 
> Cool. And thanks! I think 300dpi is right for rendering. Do you have any kind of automatic png converter from LaTeX? I know how to do it using a drag-and-drop interface, but that's a huge pain.
> 

I use ImageMagick on Linux to do the conversion: `convert -density 300 foobar.pdf foobar.png`  A few more bytes can probably be saved by then running it through an optimizer but I didn't bother.

> What we really need to do is generate those images using markdown!

Yes, or R plots with embedded Rds data to regenerate as the font, size, or style changes. Thankfully knitr will use the PDF file if it can (its available and the output is PDF) and fallback to a PNG image otherwise.

> 
> > Split many equations across multiple lines because they were too long to be printed on one line
> 
> Thanks. As you can imagine, this version targeted the html and neglected the pdf.
> 

No problem. This is actually where I started to get a version I could print through lulu.com (which then snowballed into other changes).

> > Changed the use of many non-name distributions (such as uniform) to be lower-cased. The style-guide (somewhere) states that named distributions (Poisson for example) are to be in title-case whereas the non-name distributions (including normal and uniform) should be in lowercase.
> 
> This has been totally inconsistent everywhere. I'd be just as happy lowercasing "Poisson" as then it'd match how things are rendered in Stan.
> 

Yes, consistency with the Stan casing might be good.

> > there aren't supposed to be spaces before/after the em-dash
> 
> I agree we should stick to American typesetting and spelling just to be consistent (they add spaces in the UK, which is where I learned to LaTeX and I'm still confused on this and a number of other points [like what to call the room with a toilet] 30 years later).
> 

Yup, for most of my writing I use the em-dash wrong.

> > Changed the typography of the cholesky_decompose function in equations to use \mathtt{}
> 
> I want to use `textrm` for math symbols. `texttt` should be reserved for computer program text. So we write `\textrm{square}` for the mathematical function and `\texttt{square}` when talking about the Stan function. Then it's tricky as to which symbol to use on the left-hand side of a definition, because mathematical functions don't take NaN inputs or +/- infinity. So I don't know what to do here, so I'm not too bothered by changes as long as they're consistent.

Consistency was the goal. Maybe:
* `\operatorname{}` for math functions (which is how `\log` and `\exp` work behind the scenes)
* `\texttt{}` for Stan functions that are referenced in the surrounding text
* lowercase `\textrm{}` for distributions such as normal, poisson, uniform, etc.

bob-carpenter:
We need \textrm to typeset sequences of letters as text.  \mathrm does its own thing and is not what we want, even for a single letter (as the letter baseline seems to vary).

I like the Lucida fonts because they're typographically sensible.  Lucida Bright, Lucida Console, Lucida Sans, and Lucida Math fonts form a coherent family that matches in terms of weight ("color" in the parlance of typography), x-height, and em-width so they look good run together.

Knuth's Computer Modern, the default for LaTeX, is great work for an amateur, and does have matching math and text fonts.  Courier doesn't match at all, though, as a typewriter font.  Nor do the sans serif ones.  So the docs wind up looking like ransom notes (typographer hyperbole for mismatched fonts).  If you add another font than Computer Modern, same problems with other fonts.

I don't know a good open access version of the Lucida fonts.  It might be possible to do something reasonable with Palatino.  Somehow the tufte format gets Lucida Console typewriter fonts in and pairs them with Palatino in the LaTeX version.

enbrown:
One thing that needs to be worked around is the incomplete understanding of embedded text-mode and markup in MathJax. So any embedded underscores in `\texttt{}`, `\textrm{}`, and `\mbox{}` work differently in LaTeX and MathJax. See https://mc-stan.org/docs/2_19/stan-users-guide/soft-k-means.html for a current example (with `\mbox{}` which will be replaced by `\textrm{}` or `\texttt{}`).

MathJax only (accurately) processes math LaTeX markup. So equations with embedded `\textrm{}` or `\texttt{}` commands, which switch back to text mode, cannot have markup in them (because its not understood by MathJax). Usually this isn't an issue, except that an underline (`_`) is a math-mode subscript command which needs to be escaped in real LaTeX in these commands but that necessary escaping it isn't understood in MathJax and so the backslash is visible. This is why the clustering chapter has equations with `diag_matrix` in the PDF version and `diag\_matrix` in the HTML version.

So something like `$a = b + \texttt{some_variable} + c$` can't be compiled by LaTeX (the underscore in `some_variable` causes an error), while the correct `$a = b + \texttt{some\_variable} + c$` which compiles then outputs the wrong string in HTML with MathJax (there is an extra backslash before the underscore.)

Here's a work-around (ugly but works): `$a = b + \textt{some}\mathtt{\_}\texttt{variable} + c$`. Here, the `\texttt{some}` and `\texttt{variable}` preserve any embedded spacing as desired and the `\mathtt{\_}` results in a console (fixed-width) underscore in both LaTeX/PDF and MathJax/HTML (and remains in math-mode which both LaTeX and MathJax treat the same).

bob-carpenter:
> On Jul 9, 2019, at 2:43 PM, Eric N. Brown <notifications@github.com> wrote:
> 
> One thing that needs to be worked around is the incomplete understanding of embedded text-mode and markup in MathJax. So any embedded underscores in \texttt{}, \textrm{}, and \mbox{} work differently in LaTeX and MathJax. See https://mc-stan.org/docs/2_19/stan-users-guide/soft-k-means.html for a current example (with \mbox{} which will be replaced by \textrm{} or \texttt{}).

How about:

* how about \texttt{foo{\textunderscore}bar}?

* or some kind of nested verbatim?  

* or just continuing to use \mbox?


> MathJax only (accurately) processes math LaTeX markup. So equations with embedded \textrm{} or \texttt{} commands, which switch back to text mode, cannot have markup in them (because its not understood by MathJax). Usually this isn't an issue, except that an underline (_) is a math-mode subscript command which needs to be escaped in real LaTeX in these commands but that necessary escaping it isn't understood in MathJax and so the backslash is visible. This is why the clustering chapter has equations with diag_matrix in the PDF version and diag\_matrix in the HTML version.

Thanks for sorting that out and explaining it.  It's always been a mystery to me.  I know LaTeX itself very well, so it just feels like markdown and mathjax are broken.

> 
> So something like $a = b + \texttt{some_variable} + c$ can't be compiled by LaTeX (the underscore in some_variable causes an error), while the correct $a = b + \texttt{some\_variable} + c$ which compiles then outputs the wrong string in HTML with MathJax (there is an extra backslash before the underscore.)
> 
> Here's a work-around (ugly but works): $a = b + \textt{some}\mathtt{\_}\texttt{variable} + c$. Here, the \texttt{some} and \texttt{variable} preserve any embedded spacing as desired and the \mathtt{\_} results in a console (fixed-width) underscore in both LaTeX/PDF and MathJax/HTML (and remains in math-mode which both LaTeX and MathJax treat the same).

This is where I really wish markdown had a macro language like LaTeX.

enbrown:
I tried the `\underscore` method and it didn't work (in MathJax). Neither did `\mbox` (which is what is currently on the documentation site.) I've updated the PR with something that "works" for both LaTeX and HTML.

bob-carpenter:
I found $a{\textunderscore}b$ renders appropriately in MathJax on Andrew's web site.

Does it behave differently for markdown?

enbrown:
Yes, they must behave differently due to bookdown/knitr/rmarkdown/pandoc/html/MathJax compared to what's expected in LaTeX. Here are 4 tests:

```
\begin{align}
  \texttt{binormal\_cdf} &= 1 \\
  \mathtt{binormal\_cdf} &= 2 \\
  \texttt{binormal}\_\texttt{cdf} &= 3 \\
  \texttt{binormal}\mathtt{\_}\texttt{cdf} &= 4 \\
  \texttt{binormal}{\textunderscore}\texttt{cdf} &= 5 \\
\end{align}
```

HTML output (screenshot from Windows Chrome):
image

PDF output (screenshot):
image

1. Didn't work in HTML (extra backslash) but the font used in PDF is different too (if zoomed in, its obvious its a longer underscore)
2. Worked in both (spacing the same as 4 despite not being a `\text??` command)
3. Doesn't work in HTML (wrong font for underscore that doesn't have enough weight and so just looks off)
4. Works in both (spacing the same as 2)
5. Doesn't work in HTML

The problem with 1 in PDF and 3 in HTML are subtle. The problem with 5 is hopefully clear. Oddly, even zooming in as much as I can, I don't see any kerning or spacing differences between 2 using just `\mathtt{}` and 4 using `\texttt{}` (for the words).

bob-carpenter:
This is why I dislike markdown so much.  It's an empircal endeavour that
makes no sense in the end.

Thanks so much for diving into this.  
Writing something like \texttt{binormal}\mathtt{\_}\texttt{cdf}
every time we want an underscore is going to be very tedious.  
At least it's only in math.

enbrown:
Yup. Thankfully with find-and-replace it wasn't too bad -- only 25 times in the entire user manual.

mitzimorris:
closing this PR,  almost all changes from this branch have been merged into master via https://github.com/stan-dev/docs/pull/126

PR #68, opened 2019-07-12
update ode doc to indicate time can be passed as var

charlesm93:
#### Submission Checklist

- [x] Builds locally 

#### Copyright and Licensing

Please list the copyright holder for the work you are submitting (this will be you or your assignee, such as a university or company): Charles Margossian, Columbia University and Aalto University.

### DIFF
--- a/src/stan-users-guide/odes.Rmd
+++ b/src/stan-users-guide/odes.Rmd
@@ -97,8 +97,8 @@ state function, although the accuracy of points near the discontinuity
 may be problematic (requiring many small steps).  An example of such a
 discontinuity is a lag in a pharmacokinetic model, where a
 concentration is going to be zero for times $0 < t < t'$ for some
-lag-time $t'$, whereas it will be nonzero for times $t \geq t'$.  As
-an example, would involve code in the system such as
+lag-time $t'$, whereas it will be nonzero for times $t \geq t'$.
+In this example example, we would use code in the system such as
 
 ```
 if (t < t_lag)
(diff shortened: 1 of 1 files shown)


charlesm93:
Note: I didn't commit the changed files I obtained when compiling the docs -- there were a lot of them. It seems to me this should not be git tracked.

charlesm93:
Also tagging @yizhang-yiz to review this PR.

src/stan-users-guide/odes.Rmd
```diff
@@ -270,7 +260,9 @@ system parameters `theta` and initial state `y0` are read in as data
 along with the initial time `t0` and observation times `ts`. The
 generated quantities block is used to solve the ODE for the specified
 times and then add random measurement error, producing observations
-`y_hat`.  Because the system is not stiff, the `rk45` solver is used.
+`y_hat`.  Because the system is not stiff, the `rk45` solver is used
+(in fact, the system being linear, the most efficient approach would
```
wds15: I am not really an expert on matrix-exponentials... but aren't some matrix-exp solvers using RK45 internally? I am not sure if this statement really holds in that generality. To my understanding it really depends on the properties of the eigensystem of the matrix whatever is best. I guess you want to lead people to the matrix exponential, but the real benefit of that one is to use the matrix exponential action which is right now not written in an efficient way due to memory issues.

In short: Would you mind making the statement less general on matrix exponentials?

charlesm93: Fixed.

Stan uses the Pade method coupled with scaling and squaring. My understanding is that the matrix exp method is more efficient because you only compute the solution at the required times, and don't need intermediate times. The exception I've seen involve stiff systems, and my conjecture is that the matrix exp is less efficient because of numerical instability (i.e. dealing with matrix whose elements have widely different orders of magnitude). But I haven't formally tested this conjecture.

PR #153, opened 2020-02-24
Clarify what gets printed in the generated quantities block

mcol:
#### Submission Checklist

- [X] Builds locally 
- [X] Declare copyright holder and open-source license: see below

#### Summary
Incorporates a suggestions from https://discourse.mc-stan.org/t/do-not-output-a-variable-from-generated-quantities/13228/5. If it can be further improved I'm happy to amend.

#### Copyright and Licensing

Marco Colombo

### DIFF
--- a/src/reference-manual/blocks.Rmd
+++ b/src/reference-manual/blocks.Rmd
@@ -565,7 +565,8 @@ After the generated quantities statements are executed, the constraints
 on the declared generated quantity variables are validated.
 
 All variables declared as generated quantities are printed as part of
-the output, unless they are declared in a nested block. For example:
+the output.  Variables declared in nested blocks are local variables,
+not generated quantities, and thus won't be printed.  For example:
 ```stan
 generated quantities {
   int a; // added to the output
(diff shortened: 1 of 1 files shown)


src/reference-manual/blocks.Rmd
```diff
@@ -565,4 +565,13 @@ After the generated quantities statements are executed, the constraints
 on the declared generated quantity variables are validated.
 
 All variables declared as generated quantities are printed as part of
```
bob-carpenter: Technically, what I wrote is correct.  Variables declared in nested blocks are local variables, not generated quantities.  Only the first variables declared at the top are considered generated quantities.  

I would rather rephrase this to add another sentnece that says that variables declared in local scopes are local variables and not generated quantities and thus won't be printed.  Same for transformed parameters and transformed data---in general, local variables are not visible beyond the block in which they're declared, which means they're not visible for output.
But I'll just merge this if it was confusing before as nobody's going to get confused by the semantic implication that

mcol:
Is this what you meant?

PR #163, opened 2020-04-07
Reworded user's guide introduction to reduce_sum and lowered grainsize expections

bbbales2:
The two edits I want to keep out of this are:

1. I changed the `reduce_sum` introduction stuff in the User's guide. It was difficult for me to read after taking a break and coming back.

2. I also lowered expectations on what grainsize = 1 might do. I don't think we do have empirical evidence that grainsize = 1 always works. It might work, sure, but at least I only ever tried it in 1 model.

The other edits I liked but they could come or go.

#### Submission Checklist

- [x] Builds locally 

#### Copyright and Licensing

### DIFF
--- a/src/functions-reference/higher-order_functions.Rmd
+++ b/src/functions-reference/higher-order_functions.Rmd
@@ -410,12 +410,12 @@ exactly. This implies that the order of summation determines the exact
 numerical result. For this reason, the higher-order reduce function is
 available in two variants:
 
-* `reduce_sum`: Compute partial sums automatically. This usually
- results in good performance without further tuning.
-* `reduce_sum_static`: For the same input, always create the same call
-graph. This results in stable numerical evaluation. This version
-requires setting a tuning parameter which controls the maximal size of
-partial sums formed.
+* `reduce_sum`: Automatically choose partial sums partitioning based on a dynamic
+ scheduling algorithm.
+* `reduce_sum_static`: Compute the same sum as `reduce_sum`, but partition
+ the input in the same way for given data set (in `reduce_sum` this partitioning
+ might change depending on computer load). This should result in stable
+ numerical evaluations.
 
 ### Specifying the Reduce-sum Function
 
@@ -437,16 +437,14 @@ partial sums. `s1, s2, ...` are shared between all terms in the sum.
 * *`f`*: function literal referring to a function specifying the
 partial sum operation. Refer to the [partial sum function](#functions-partial-sum).
 * *`x`*: array of `T`, one for each term of the reduction, `T` can be any type,
-* *`grainsize`*: recommended number of terms in each reduce call, set
-to 1 to estimate automatically for `reduce_sum` while for
-`reduce_sum_static` this determines the maximal size of the partial sums, type `int`,
+* *`grainsize`*: For `reduce_sum`, `grainsize` is the recommended size of the partial sum. For `reduce_sum_static`, `grainsize` determinse the maximum size of the partial sums, type `int`,
 * *`s1`*: first (optional) shared argument, type `T1`, where `T1` can be any type
 * *`s2`*: second (optional) shared argument, type `T2`, where `T2` can be any type,
 * *`...`*: remainder of shared arguments, each of which can be any type.
 
-### The Partial-sum Function {#functions-partial-sum}
+### The Partial sum Function {#functions-partial-sum}
 
-The partial sum function must have the following signature where the types `T`, and the
+The partial sum function must have the following signature where the type `T`, and the
 types of all the shared arguments (`T1`, `T2`, ...) match those of the original
 `reduce_sum` (`reduce_sum_static`) call.
 
--- a/src/functions-reference/higher-order_functions.Rmd
+++ b/src/functions-reference/higher-order_functions.Rmd
@@ -135,7 +135,7 @@ package MINPACK-1 [@minpack:1980].
 
 The Jacobian of the solution with respect to auxiliary parameters is
 computed using the implicit function theorem. Intermediate Jacobians
-(of the the algebraic function's output with respect to the unknowns y
+(of the algebraic function's output with respect to the unknowns y
 and with respect to the auxiliary parameters theta) are computed using
 Stan's automatic differentiation.
 
@@ -422,7 +422,7 @@ available in two variants:
 The higher-order reduce function takes a partial sum function `f`, an array argument `x`
 (with one array element for each term in the sum), a recommended
 `grainsize`, and a set of shared arguments. This representation allows
-to parallelize the resultant sum.
+parallelization of the resultant sum.
 
 <!-- real; reduce_sum; (F f, T[] x, int grainsize, T1 s1, T2 s2, ...); -->
 \index{{\tt \bfseries reduce\_sum }!{\tt (F f, T[] x, int grainsize, T1 s1, T2 s2, ...): real}|hyperpage}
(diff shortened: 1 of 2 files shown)


src/functions-reference/higher-order_functions.Rmd
```diff
@@ -437,16 +437,14 @@ partial sums. `s1, s2, ...` are shared between all terms in the sum.
 * *`f`*: function literal referring to a function specifying the
 partial sum operation. Refer to the [partial sum function](#functions-partial-sum).
 * *`x`*: array of `T`, one for each term of the reduction, `T` can be any type,
-* *`grainsize`*: recommended number of terms in each reduce call, set
-to 1 to estimate automatically for `reduce_sum` while for
-`reduce_sum_static` this determines the maximal size of the partial sums, type `int`,
+* *`grainsize`*: For `reduce_sum`, `grainsize` is the recommended size of the partial sum. For `reduce_sum_static`, `grainsize` determinse the maximum size of the partial sums, type `int`,
```
wds15, line 440: "determinse" ?

src/stan-users-guide/parallelization.Rmd
```diff
@@ -28,25 +28,21 @@ law](https://en.wikipedia.org/wiki/Amdahl's_law).
...
 
 `g(x1) + g(x2) + ...`
 
-In probabilistic modeling this comes up when there are $N$
-conditionally independent terms in a likelihood. Because of the
-conditional independence, these terms can be computed in parallel. If
-dependencies exist between the terms, then this isn't possible. For
-instance, in evaluating the log density of a Gaussian process then
-summation of independent terms isn't applicable.
+The `reduce_sum` function is a tool for automatically parallelizing these
```
wds15, line 39: I avoided to speak directly of `reduce_sum`, but rather speak of reduce summation. The actual implementations are then `reduce_sum` and `reduce_sum_static`. Does that logic make sense to you?

src/stan-users-guide/parallelization.Rmd
```diff
@@ -261,35 +257,41 @@ model {
...
-figure out an optimal grainsize, think about how many terms are in the
-summation and on how many cores the model should run. If there are `N`
+In `reduce_sum`, `grainsize` is a recommendation on how to partition
+the work in the partial sum into smaller pieces. A `grainsize` of 1
+leaves this entirely up to the internal scheduler. Ideally this will be
+efficient, but there are no guarantees.
+
+In `reduce_sum_static`, `grainsize` is an upper limit on the worksize.
+Work will be split until all partial sums are just smaller than `grainsize`
+(and the split will happen the same way every time for the same data).
```
wds15, line 278: I would rather speak of "the same data" of "the same input" to make clear that everything is meant (data and parameters).

src/functions-reference/higher-order_functions.Rmd
mitzimorris: determines?

src/stan-users-guide/parallelization.Rmd
```diff
@@ -28,25 +28,21 @@ law](https://en.wikipedia.org/wiki/Amdahl's_law).
...
+calculations.
 
 For efficiency reasons the reduce function doesn’t work with the
-element-wise evaluated function `g`, but instead requires the partial
-sum function ```f: U[] -> real```, where ```f``` computes the partial
-sum corresponding to a slice of the sequence ```x``` passed in. Due to the
+element-wise evaluated function `g`, but instead the partial
+sum function `f: U[] -> real`, where `f` computes the partial
+sum corresponding to a slice of the sequence `x` passed in. Due to the
 the associativity of the sum reduction it holds that:
```
mitzimorris, line 46: copy edit - "the the"!

src/stan-users-guide/parallelization.Rmd
```diff
@@ -261,35 +257,41 @@ model {
 
 ### Picking the Grainsize {#reduce-sum-grainsize}
 
-The `grainsize` is a recommendation on how large each piece of
+For `grainsize` is a recommendation on how large each piece of
```
mitzimorris, line 260: "For" is ungrammatical

src/stan-users-guide/parallelization.Rmd
wds15: Ah, now I see your de-emphasis of my "let grainsize equal to 1" recommendation. I would actually really prefer to more strongly recommend people to just use grainsize = 1 and that's it. From what I have seen, this is well justified.

src/stan-users-guide/parallelization.Rmd
```diff
@@ -261,35 +257,41 @@ model {
...
 parallel work is (how many terms it contains). When using the
 non-static version, it is recommended to choose 1 as a starting
 point as automatic aggregation of partial sums are performed. However,
 for the static version the `grainsize` defines the maximal size of the
-partial sums, e.g. the static variant will split the input sequence
-until all partial sums are just smaller than `grainsize`. Therefore,
-for the static version it is more important to select a sensible
-value. The rational for choosing a sensible `grainsize` is based on
+partial sums, e.g.
```
mitzimorris, line 265: ???

bbbales2: I forgot to delete this paragraph

mitzimorris:
following your changes - perhaps too soon?
I like these edits, FWIW

src/stan-users-guide/parallelization.Rmd
bbbales2: Yeah that's how I'd like it to be, but I don't know if it is gonna turn out that way.

Honestly I'm not entirely sure about the behavior of grainsize with the auto partitioner. I'm happy to say, 'If nothing else, we recommend setting this to 1', but I don't see why we'd imply you shouldn't change this.

If you really shouldn't change this, it shouldn't be an option!

What are the rules on this with the auto partitioner. Is it something like the partitioner won't split stuff below this size? So it's sorta the opposite of the simple partitioner in that way?

wds15: TBB doc for auto_partitioner:

> A loop template with an auto_partitioner attempts to minimize range splitting while providing ample opportunities for work-stealing.

> The range subdivision is initially limited to S subranges, where S is proportional to the number of threads specified by the task_scheduler_init or task_arena. Each of these subranges is not divided further unless it is stolen by an idle thread. If stolen, it is further subdivided to create additional subranges. Thus a loop template with an auto_partitioner creates additional subranges only when necessary to balance load.

> TIP
> When using auto_partitioner and a blocked_range for a parallel loop, the body may receive a subrange larger than the grain size of the blocked_range. Therefore do not assume that the grain size is an upper bound on the size of a subrange. Use simple_partitioner if an upper bound is required.

So the auto_partitioner splits according to how many threads you have available and it still provides possibilities for work stealing if that is possible. This will likely lead to larger blocks rather than smaller blocks for Stan programs.

Now, for Stan's autodiff the strategy to make it as large as possible is - that is what my hunch is - not always great. Once the work chunk size blows the CPU cache size, then this is really bad news for the reverse mode AD which is super memory intense.

Still, I would think that we confuse users a lot if we encourage to always tune it.

So we can settle on 'If nothing else, we recommend setting this to 1' - and this is where we one should start. People should concentrate on their problems, not on the grainsize.

bbbales2: > So we can settle on 'If nothing else, we recommend setting this to 1'

Agreed. I'll make changes to that effect here in a few minutes.

bbbales2:
@wds15, @mitzimorris ready again

src/stan-users-guide/parallelization.Rmd
bbbales2: I just added them both here. I wanted to say the function names explicitly to avoid talking about a reduce, or parallel for, or map or any of that stuff.

src/stan-users-guide/parallelization.Rmd
```diff
@@ -222,15 +220,15 @@ accordingly with `start:end`. With this function, reduce summation can
...
                      grainsize,
                      x, beta);
 ```
 
-The reduce summation facility automatically breaks the sum into roughly `grainsize` sized pieces
-and computes them in parallel. `grainsize = 1` specifies that the grainsize should
-be estimated automatically. The final model looks as:
+The reduce summation facility automatically breaks the sum into pieces
+and computes them in parallel. `grainsize = 1` specifies that the
+`grainsize` should be estimated automatically. The final model looks as:
```
mitzimorris, line 231: copy edit: "the final model is:"

bbbales2:
@wds15 All ready if yah wanna merge this up to the other pull and finish whatever needs to be done there.

Issue #160, opened 2020-03-22
Missing plot in ODE section of user's guide


bbbales2:
#### Summary:
The plot should be here: https://mc-stan.org/docs/2_22/stan-users-guide/measurement-error-models.html

#### Description:
Here's some R code that should produce a plot near what I think this was meant be:

```
library(tidyverse)
library(ggplot2)
library(deSolve)

f = function(t, y, parms) {
  return(list(c(y[2], -y[1] - 0.15 * y[2])))
}

out = ode(c(y0 = 1.0, y1 = 0.0), seq(1.0, 10.0, length = 10), f) %>%
  as_tibble() %>%
  gather(state, y, y0, y1) %>%
  mutate(y_noise = rnorm(length(y), y, 0.1))

out_highres = ode(c(y0 = 1.0, y1 = 0.0), seq(1.0, 10.0, length = 1000), f) %>%
  as_tibble() %>%
  gather(state, y, y0, y1)

out %>%
  ggplot(aes(time, y_noise)) +
  geom_point(aes(color = state)) +
  geom_path(data = out_highres, aes(time, y, group = state, color = state),
            linetype = "dashed")
```

Presumably it'll need formatted to be like other plots in the manual.

(edit: added missing libraries)

#### Current Version:
v2.22.0

torkar:
@bbbales2 , I'm not 100% sure but I think I fixed this on GitHub last week. Can you check?

PR #164, opened 2020-04-09
Footnote

breckbaldwin:
#### Submission Checklist

- [x] Builds locally 
- [x ] Declare copyright holder and open-source license: see below

#### Summary
Initial effort on footnote additions to user guide. Only one footnote done for 'Linear Regression' section. 

Added in a local makefile 'docs/src/stan-users-guide' that reflects how docs are built in 'docs/build.py', make pdf is not working with error:

Quitting from lines 5423-5424 (_main.Rmd) 
Error in magick_image_readpath(enc2native(path), density, depth, strip) : 
  R: UnableToOpenBlob `_main_files/figure-latex/unnamed-chunk-11-1.png': No such file or directory @ error/blob.c/OpenBlob/2761
Calls: <Anonymous> ... in_dir -> plot_crop -> <Anonymous> -> magick_image_readpath

#### Copyright and Licensing

Columbia University


bob-carpenter:
I haven't seen that error before, but the clue is here:

> Quitting from lines 5423-5424 (_main.Rmd)

The `_main.Rmd` is the Rmd file generated by bookdown from all the little .Rmd files specified in `_bookdown.yml`.  You should be able to navigate to line 5423 of that file and see where the error is coming from.

Can you build the develop branch?

src/stan-users-guide/regression.Rmd
```diff
@@ -25,24 +27,15 @@ $$
 y_n \sim \operatorname{normal}(\alpha + \beta X_n, \, \sigma).
 $$
 
-This latter form of the model is coded in Stan as follows.
+This latter form of the model is coded in Stan as follows [linear_regression_1](https://github.com/stan-dev/docs/tree/footnote/src/stan-users-guide/linear_regression_1)
```
bob-carpenter, line 30: 1. `footnote/src/stan-users-guide/linear_regression_1` should be part of this PR.

2.  Shouldn't this be pointing to what is called `linear_regression_1/regression_1.stan` below?

3.  This won't be gramatical when rendered---it'll say 

> This latter form of the model is coded in stan as follows [linear_regression_1](foo.html)

It's missing connectors and a final sentence period.  I'd suggest coding it so it renders as:

> This latter form of the model is coded in stan as follows (see [source code](foo.html)).

You can look at the source of this comment to see how it needs to be coded.

src/stan-users-guide/linear_regression_1/regression_1.stan
```diff
@@ -0,0 +1,13 @@
...
+  vector[N] x;
+  vector[N] y;
+}
+parameters {
+  real alpha;
+  real beta;
+  real<lower=0> sigma;
+}
+model {
+  y ~ normal(alpha + beta*x, sigma);
```
bbbales2, line 12: Inconsistent!

src/stan-users-guide/regression.Rmd
```diff
@@ -25,24 +27,15 @@ $$
...
-  real alpha;
-  real beta;
-  real<lower=0> sigma;
-}
-model {
-  y ~ normal(alpha + beta * x, sigma);
-}
-```
 
+```{r echo=FALSE}
```
bbbales2, line 33: I'm still out on including files. If I were writing this, the fact that I can now link to a file makes it so that the only thing I'd put in the text itself would be a minimal code snippet:

```
model {
  y ~ normal(alpha + beta * x, sigma);
}
```

And then leave the full code example for the link.

src/stan-users-guide/linear_regression_1/regression_1.stan
bob-carpenter: Follow the guidelines and use spaces around all operators, including `=` and `*` here. 

Also, this is going to need to have priors if it's going to be the kind of reference model we can't point people to.  We're going to have to rewrite the manual so that it starts with more complicated best practices so that we don't publish models we don't want people to use.

Or, this is going to have to have a huge warning spattered across it saying "DO NOT USE---FOR PEDAGOGICAL PURPOSES ONLY; SEE <PATH> FOR REPLACEMENT"

bbbales2: > Also, this is going to need to have priors

Good catch.

breckbaldwin:
> I haven't seen that error before, but the clue is here:
> 
> > Quitting from lines 5423-5424 (_main.Rmd)
> 
> The `_main.Rmd` is the Rmd file generated by bookdown from all the little .Rmd files specified in `_bookdown.yml`. You should be able to navigate to line 5423 of that file and see where the error is coming from.

I built off of branch docs-2-22 and still got the error. See below where I can build the problem files stand alone. 

The offending line is from  docs/src/stan-users-guide/latent-discrete.Rmd:

```{r include=TRUE, fig.align="center", fig.cap=c("Analytical change-point posterior"), echo=FALSE}
knitr::include_graphics("./img/change-point-posterior.png", auto_pdf = TRUE)
```
I commented out the .Rmd file include _bookdown.yml, got another error at:

```{r fig.cap="Simple harmonic oscillator trajectory"}
knitr::include_graphics("img/sho-ode-trajectory.png", auto_pdf = TRUE)
```
from docs/src/stan-users-guide/odes.Rmd

I can build fine with the two above files commented out of _bookdown.yml. 

I also can build the offending files independently using the same commands so this is looking more difficult. 

I am just going to keep dissecting the code and checking versions until I can get it sorted.  

No action needed on Ben/Bob's part unless something clicks from the above.

andrewgelman:
Hi, Bob.  I don't see why a linear regression with flat priors is "NOT FOR USE".  We have lots of models like this in the manual, no?

To step back a moment:  I agree that flat priors are not a good default choice, but in particular applications it can be fine to use a flat prior.  My goal with these "footnotes" was to have simple models and not have to go into lots of details that would make each model into a "case study."

bob-carpenter:
I'm pinging @andrewgelma because I don't know if he sees these otherwise.

> Hi, Bob.  I don't see why a linear regression with flat priors is "NOT FOR USE".

I'm just channeling @avehtari, who requested I change my last pull request to not include a model with no prior.  I thought the current policy was all models get priors.

There's also a question of how efficient these models should be.  That's *not* the most efficient way to write linear regression in Stan.  We have an actual linear regression lpdf that's more efficient.  I was stumped trying to balance pedagogy and good practice in rewriting our example models.  I just gave up because of all the unclear competing demands (pedagogy, best stats practices, and efficiency all pull in different directions).

bob-carpenter:
> No action needed on Ben/Bob's part unless something clicks from the above.

Does `img/change-point-posterior.png` exist?  I see `change-point-posterior.pdf` in my directory on the current develop branch.

P.S.  The `./` isn't doing anything at the start of a path

breckbaldwin:
> Does `img/change-point-posterior.png` exist? I see `change-point-posterior.pdf` in my directory on the current develop branch.
I added the .png and changed the reference in the .Rmd in thinking it might need it a while back and forgot. Doesn't fix the bug. 
> P.S. The `./` isn't doing anything at the start of a path
Yep.

breckbaldwin:
So this is more busted than I thought. Current doc is not showing images at:

https://mc-stan.org/docs/2_22/stan-users-guide/change-point-section.html

Look to Figure 1. 

The pdf is fine.

bbbales2:
@breckbaldwin Ooooo, really? I ran into something like this as well: https://github.com/stan-dev/docs/issues/160

If that's happened throughout the manual we needta get that fixed. Is this a pdf vs. html thing? Can you tell?

bbbales2:
> I'm just channeling @avehtari, who requested I change my last pull request to not include a model with no prior. I thought the current policy was all models get priors.

I agree with adding in priors. It's easy, and maybe it makes people think through having a prior when maybe they woulda forgotten one or whatever. Also it makes it easier to code review lol -- no worrying about if it's reasonable or not to have a prior -- we just gotta have one.

breckbaldwin:
> @breckbaldwin Ooooo, really? I ran into something like this as well: #160
> 
> If that's happened throughout the manual we needta get that fixed. Is this a pdf vs. html thing? Can you tell?

#160 looks like the same issue, it was not building either and I had pulled it out--it is in odes.Rmd. 

I'll pick up this on #160, we can keep this about the footnotes.

andrewgelman:
We should discuss at the next Stan meeting.  I'm fine with adding priors, but there's the question of what priors to add etc!

bob-carpenter:
That's also an issue for fixing all the bugs models with those normal(0, 1e4) and gamma(1e-3, 1e-3) priors.   There's the recommended prior Wiki, but it's a bit vague and general.

andrewgelman:
Good point.

I guess the simplest approach would just be to use things like N(0,1) and N+(0,1) priors wherever possible and then just have a general note somewhere in the users guide, or whatever we write to document the "footnotes," to explain that we are using these as default priors with the understanding that users will either put their parameters on unit scale or else scale their priors appropriately.  We can give a couple of examples.  I can write that bit.

So, for now, my proposal is that, except where we decide otherwise, we use normal(0,1) priors for everything (which implicitly is normal+(0,1) for parameters that are constrained to be positive).

Andrew

P.S.  It's great how we can have these discussions and make progress!

bob-carpenter:
Here's a good version of the first BUGS model, which I wrote to be efficient.  Note that it does some data type gymnastics in transformed data.    The original is in: http://www.openbugs.net/Examples/Rats.html

I set the prior scales to 4 * the actual values from a previous fit, which seems roughly "weakly informative".  

```
data {
  int<lower = 0> N;
  int<lower = 0> T;
  vector[T] x;
  matrix[N, T] y;
}
transformed data {
  real x_bar = mean(x);
  vector[T] x_ctr = x - x_bar;
  vector[N] y_tr[T];
  for (t in 1:T)
    y_tr[t] = y[ , t];
}
parameters {
  vector<offset = mu_alpha, multiplier = sigma_alpha>[N] alpha;
  vector<offset = mu_beta, multiplier = sigma_beta>[N] beta;
  real<mulitplier = 5> mu_alpha;
  real<mulitplier = 5> mu_beta;
  real<lower = 0> sigma_y;
  real<lower = 0> sigma_alpha;
  real<lower = 0> sigma_beta;
}
model {
  mu_alpha ~ normal(0, 1000);
  sigma_alpha ~ normal(0, 60);
  alpha ~ normal(mu_alpha, sigma_alpha);
  mu_beta ~ normal(0, 25);
  sigma_beta ~ normal(0, 2);
  beta ~ normal(mu_beta, sigma_beta);
  sigma_y ~ normal(0, 25);
  for (t in 1:T)
    y_tr[t] ~ normal(alpha + beta * x_centered[t], sigma_y);
}
generated quantities {
  real alpha0 = mu_alpha - mu_beta * x_bar;
}
```

breckbaldwin:
> I guess the simplest approach would just be to use things like N(0,1) and N+(0,1) priors wherever possible and then just have a general note somewhere in the users guide, or whatever we write to document the "footnotes," to explain that we are using these as default priors with the understanding that users will either put their parameters on unit scale or else scale their priors appropriately. We can give a couple of examples. I can write that bit. So, for now, my proposal is that, except where we decide otherwise, we use normal(0,1) priors for everything (which implicitly is normal+(0,1) for parameters that are constrained to be positive).

The footnotes have data so these scale issues can be made clear. The footnotes can have narrative as well, perhaps as comments in the source or maybe their own markdown. Maybe there would be non-unit scaled data that gets scaled and fit and a version that scales the priors to make it absolutely clear what is going on.

breckbaldwin:
> There's also a question of how efficient these models should be. That's _not_ the most efficient way to write linear regression in Stan. We have an actual linear regression lpdf that's more efficient. I was stumped trying to balance pedagogy and good practice in rewriting our example models. I just gave up because of all the unclear competing demands (pedagogy, best stats practices, and efficiency all pull in different directions).

I suggest we show the most efficient model with best practices.  For early chapters break it down into easier versions for ease of understanding. I'd consider making vectorized operations explicit with loops, data specified in transformed data {} and explain the ~ and += conventions.

andrewgelman:
The footnotes actually use fake data, so the priors won't be a problem there.  It will just be an issue when users take them as example code and have to figure out what to do with the priors.

andrewgelman:
It's also ok to have multiple models in the same directory, even with names such as linear_regression_naive_coding and linear_regression_efficient_coding

bob-carpenter:
> I suggest we show the most efficient model with best practices.

We're going to have to make some compromises.

The issue I struggle with is balancing pedagogy with efficiency.  Best practices themselves are tricky, as best practice for maintenance is easy to read code, whereas best practice for efficiency is highly unrolled, special-function laden and matrix-oriented.

And while it'd be great if the compiler was a magician, that hasn't happened everywhere for C++, so it's not going to happen everywhere for Stan.

> It's also ok to have multiple models in the same directory,

Can we just name the directories after the chapters of the manual?  Or is there going to be one footnotes directory per model, perhaps with multiple implementations of that model.  By model here, I mean what winds up being the function from data to output draws.  I'd rather not have programs that expect different data in the same directories as then the input, simulations, etc. all have to multiply.  

But I haven't actually done this, so I don't know how it'll play out.

Or is there some other organization that makes more sense?

andrewgelman:
My inclination would be to have each "footnote" in a mini-directory that has a descriptive name (e.g., SImple_Linear_Regression), rather than being tied to chapters of the manual.  My concern about being tied to chapters of the manual is that I could imagine the manual evolving with chapters split in two, etc.

I was assuming that a "footnotes" directory would have, at minimum, one .stan file, one .R file, and one .py file, but it could have other things, such as data files (I'd prefer to do things with fake data that are simulated directly in the code), but I guess there might be some examples where we'd want to use real data) or additional .stan files if, for example, we want to illustrate multiple ways of coding the same model.

But I'm obviously not a Github expert; this is all just my naive thinking.

avehtari:
First, as mentioned in the discussion I prefer to have explicit proper prior for everything. The reasons are 1) for new users it's much easier for more complete code, e.g. it's much easier to modify prior parameter values than figure out from other source what is the recommended prior, that you need to add something to parameters block and model block and have lower=0 in some cases and not use uniform etc., 2) we are developing prior sensitivity and data-prior conflict diagnostics which is much easier to implement for proper priors.

> I guess the simplest approach would just be to use things like N(0,1) and N+(0,1) priors wherever possible and then just have a general note somewhere in the users guide, or whatever we write to document the "footnotes," t

Probably that general note need to be linked from each footonote, as people find them by googling and never see the general not otherwise.

>  to explain that we are using these as default priors with the understanding that users will either put their parameters on unit scale or else scale their priors appropriately.  We can give a couple of examples. 

Yes!

> So, for now, my proposal is that, except where we decide otherwise, we use normal(0,1) priors for everything (which implicitly is normal+(0,1) for parameters that are constrained to be positive).

I think it would be useful to mark it explicitly when we have decided otherwise, and maybe eventually we can have links to references discussing priors and how to think about the prior parameters for all different types of models.

bob-carpenter:
Thanks for clarifying, Aki.  The most convincing argument for me is that we can't do things like prior predictive checks or SBC without priors.  But I didn't understand the technical implications of [X(i'an)'s recent blog post](https://xianblog.wordpress.com/2020/04/10/mea-culpa-2/).

> maybe eventually we can have links to references discussing priors and how to think about the prior parameters for all different types of models.

It would be great to merge the prior recommendations into the user's guide.  Or into its own standalone book given that it's such a deep dive.

> But I'm obviously not a Github expert; this is all just my naive thinking.

There's nothing GitHub specific about how we organize the files other than that users will be able to navigate GitHub we pages to navigate and view them.

andrewgelman:
Given that the footnotes will be linked to the user's guide, and we'll need to have a short document explaining the footnotes, I guess you're right!

src/stan-users-guide/regression.Rmd
```diff
@@ -25,24 +27,15 @@ $$
...
+library(knitr)
+library(readr)
+stan_file = read_file('linear_regression_1/regression_1.stan')
+writeLines(stan_file)
+```
```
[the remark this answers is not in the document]
breckbaldwin: Changed to print_file() from before-chapter.R--pushed

Directory name 'linear_regression_1' reflects heading name for sub-section, '## Linear Regression' . This seems the easiest way for readers to track what dir they want. The Regression.Rmd covers a range of topics, e.g., QR Decomposition which would get a directory 'qr_decomposition_n' where n increments based on order in the chapter. 

If there is need for multiple implementations in the directory I think they can whatever file name they want. Looking ahead in the comments I think I'd like to to have something like
'linear_regression_naive_coding.stan' and 'linear_regression_efficient_coding.stan' for this directory. The interface programs that call the stan code may or may not be 1-1 so 'linear_regression.R' may have calls to both .stan models. 

Not sure I like the numeric suffix but they function like figure numbers which I think is helpful.

src/stan-users-guide/regression.Rmd
breckbaldwin: > 

> 1. `footnote/src/stan-users-guide/linear_regression_1` should be part of this PR.
Did I screw up? Do you not see:https://github.com/stan-dev/docs/tree/footnote/src/stan-users-guide/linear_regression_1 ? 

> 2. Shouldn't this be pointing to what is called `linear_regression_1/regression_1.stan` below?
No, we want to point to the directory that contains code that has interface programs that can run the .stan program. If we link straight to the .stan file the other programs are not obvious.

> 3. This won't be gramatical when rendered---it'll say
> 
> > This latter form of the model is coded in stan as follows [linear_regression_1](foo.html)
> 
> It's missing connectors and a final sentence period. I'd suggest coding it so it renders as:
> 
> > This latter form of the model is coded in stan as follows (see [source code](foo.html)).
> 

How about something like: This latter form of the model is coded in stan as follows (see [interface/source code](foo.html)).

src/stan-users-guide/linear_regression_1/regression_1.R
```diff
@@ -0,0 +1,26 @@
+#library("rstan")
+library("cmdstanr")
+set_cmdstan_path("~/cmdstan-2.22.1/") # set to your local install of cmdstan
```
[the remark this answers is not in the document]
breckbaldwin: Done

src/stan-users-guide/linear_regression_1/regression_1.R
```diff
@@ -0,0 +1,26 @@
...
+
+#Run from command line: Rscript regression_1.R
```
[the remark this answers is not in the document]
breckbaldwin: We need the library("cmdstanr") to run. 

You don't want instructions for command line execution?

src/stan-users-guide/regression.Rmd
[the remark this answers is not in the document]
bob-carpenter: I think Andrew makes a good point that the footnotes can stay in place while chapters, sections, etc. get reorganized.  

> My inclination would be to have each "footnote" in a mini-directory that has a descriptive name (e.g., SImple_Linear_Regression), rather than being tied to chapters of the manual.  My concern about being tied to chapters of the manual is that I could imagine the manual evolving with chapters split in two, etc.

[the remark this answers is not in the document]
bob-carpenter: `Simple_Linear_Regression` uses both camel case and underscore separators.  We don't need both.  So choose either `SimpleLinearRegression` or `simple_linear_regression`.  I strongly prefer the latter.  I like `simple-linear-regression` with hyphens even more.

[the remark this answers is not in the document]
breckbaldwin: Ok, I'll drop the suffix numbering scheme and keep directory names descriptive. 

I'd like one naming scheme that works across .stan, .R and .py programs based on legal variable names. Stan/R doesn't allow for '-' and camel case does not seem adopted in R so that leaves us with '_'.  Everyone ok with that?

src/stan-users-guide/linear_regression_1/regression_1.R
```diff
@@ -0,0 +1,26 @@
+#library("rstan")
```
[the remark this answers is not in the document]
breckbaldwin: Done

src/stan-users-guide/regression.Rmd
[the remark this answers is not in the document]
bob-carpenter: Yes, I'm OK with that.

P.S.  `-` works in file names in all of those languages.  As do file names starting with integers.  Where we've run into problems is when interfaces automatically promote file names to identifiers.

src/stan-users-guide/linear_regression_1/regression_1.py
```diff
@@ -0,0 +1,23 @@
+import os
+from cmdstanpy import cmdstan_path, CmdStanModel
+import numpy as np
+import fileinput
+import sys
```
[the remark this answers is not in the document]
breckbaldwin: removed

src/stan-users-guide/linear_regression_1/regression_1.py
```diff
@@ -0,0 +1,23 @@
...
+# Run from command line: Python regression_1.py
+
+alpha = 2
+beta = 3
+sigma = 5
+N = 100
+x = np.random.uniform(size = N)
+y = np.random.normal(size = N, loc=alpha + beta * x, scale = sigma)
```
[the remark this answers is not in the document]
breckbaldwin: I went and ran pylint on the file and got their perfect score. I think pylint implements https://www.python.org/dev/peps/pep-0008/. 

Have a look, lots of ALL_CAPS for python's unofficial constants.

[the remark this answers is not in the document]
bbbales2: ```size = N``` vs ```loc=alpha + beta * x```

Hmm, not impressed with their linter settings. There should be spaces around the equals signs.

src/stan-users-guide/regression.Rmd
breckbaldwin: I am avoiding doing rewrite for now just to keep the size of the task manageable.  What you say makes sense.

src/stan-users-guide/linear_regression_1/regression_1.py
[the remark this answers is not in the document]
bbbales2: Space around spaces falls under this: https://www.python.org/dev/peps/pep-0008/#other-recommendations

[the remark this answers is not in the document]
bbbales2: Spaces around equals I mean lol.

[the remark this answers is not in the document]
breckbaldwin: I pushed a very different version, have I screwed something up? 
https://github.com/stan-dev/docs/blob/footnote/src/stan-users-guide/linear_regression_1/regression_1.py

For example 

N = 100
X = numpy.random.uniform(size=N)
Y = numpy.random.normal(size=N, loc=ALPHA + BETA * X, scale=SIGMA)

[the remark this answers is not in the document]
bbbales2: Oh I was looking at outdated code. But yeah I don't like the all caps code. It says in their pep to put spaces around equals -- I wonder why the linter likes this other thing?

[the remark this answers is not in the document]
breckbaldwin: Ok, I moved it back to lowercase variables, looks way better IMHO.  pylint complains, pycodestyle does not.

andrewgelman:
Underscores make sense to me, given other conventions we have in coding.

breckbaldwin:
> First, as mentioned in the discussion I prefer to have explicit proper prior for everything.

I have added priors to two .stan models and incorporated suggested changes. 


> My inclination would be to have each "footnote" in a mini-directory that has a descriptive name (e.g., SImple_Linear_Regression), rather than being tied to chapters of the manual.

Done, directory is named 'simple_linear_regression' and linked from the Regression.Rmd chapter.

> I was assuming that a "footnotes" directory would have, at minimum, one .stan file, one .R file, and one .py file, but it could have other things, such as data files (I'd prefer to do things with fake data that are simulated directly in the code), but I guess there might be some examples where we'd want to use real data) or additional .stan files if, for example, we want to illustrate multiple ways of coding the same model.

We have run.py, run.R and a README.txt as well as an output/ directory. There are two .stan files to run, regression_naive.stan and regression_efficient.stan. Both are called from the run.*

As checked in under the footnote branch the book won't build because I don't have the relevant .png files in img/. They are in master and I'll merge but wanted to make sure I did it correctly so I'll ask at the meeting.

src/stan-users-guide/simple_linear_regression/output/.gitignore
```diff
@@ -0,0 +1,5 @@
+#OSX
```
bob-carpenter, line 1: I don't think .gitignore should be in subdirectories, just the top-level directory.

src/stan-users-guide/simple_linear_regression/regression_efficient.stan
```diff
@@ -0,0 +1,19 @@
+data {
+  int<lower=0> N;
+  vector[N] x;
+  vector[N] y;
+}
+
+parameters {
+  real alpha_hat;
```
bob-carpenter, line 8: Parameters should not be suffixed `_hat`.  The notation alpha-hat is used for a point estimate of the parameter `alpha`, so the parameter is just `alpha`. (This gets subtle when you define things like y-hat in the model, but my recommendation is to never call anything hat in the model.)

src/stan-users-guide/simple_linear_regression/regression_efficient.stan
```diff
@@ -0,0 +1,19 @@
+data {
+  int<lower=0> N;
```
bob-carpenter, line 2: Space around all operators.  That's `=` and `*` in this example.

src/stan-users-guide/simple_linear_regression/regression_efficient.stan
```diff
@@ -0,0 +1,19 @@
...
+
+parameters {
+  real alpha_hat;
+  real beta_hat;
+  real<lower=0> sigma_hat;
+}
+
+//Just vectorized right now. 
+model {
+  alpha_hat ~ normal(5,10);
```
bob-carpenter, line 15: @andrewgelman suggested making these all unit scale because we're just simulating data.  So that'd be
`normal(0, 1)` for all these.

bob-carpenter, line 15: Spaces after commas always.

src/stan-users-guide/simple_linear_regression/regression_naive.stan
```diff
@@ -0,0 +1,20 @@
+data {
```
bob-carpenter, line 1: Same comments as in other model.  I'm not convinced we need a model with a loop as a baseline, but that's a judgement call.

src/stan-users-guide/simple_linear_regression/regression_efficient.stan
```diff
@@ -0,0 +1,19 @@
+data {
```
bob-carpenter, line 1: This shouldn't be called "efficient" because it's really the baseline.  The more efficient thing to do is to use the glm function, which would have slightly different data types.

src/stan-users-guide/simple_linear_regression/run.R
```diff
@@ -0,0 +1,32 @@
+library("cmdstanr")
+
+#set_cmdstan_path(paste(Sys.getenv("HOME"),"/.cmdstanr/cmdstan",sep=''))
+
+# Run from command line: Rscript regression_1.R
```
bob-carpenter, line 5: Don't document the language in code.

src/stan-users-guide/simple_linear_regression/run.R
```diff
@@ -0,0 +1,32 @@
+library("cmdstanr")
+
+#set_cmdstan_path(paste(Sys.getenv("HOME"),"/.cmdstanr/cmdstan",sep=''))
```
bob-carpenter, line 3: Remove rather than comment out code.

src/stan-users-guide/simple_linear_regression/run.R
```diff
@@ -0,0 +1,32 @@
...
+
+# Run from command line: Rscript regression_1.R
+# If running from RStudio remember to set the working directory 
+# >Session>Set Working Directory>To Source File Location
+
+#generate data
```
bob-carpenter, line 9: Data should be generated from the priors.

src/stan-users-guide/simple_linear_regression/run.R
```diff
@@ -0,0 +1,32 @@
...
+N <- 100
+x <- runif(N, 0, 10)
+y <- rnorm(N, alpha + beta * x, sigma)
+
+stan_data <- list(N = N, x = x, y = y)
+
+print(sprintf("generating parameters are: alpha=%.1f, beta=%.1f, sigma=%.1f, N=%d",
+              alpha,beta,sigma,N))
+
+model <- cmdstan_model('regression_naive.stan')
```
bob-carpenter, line 22: Choose either `'` or `"` for quoting strings and stick with it.  The first string in this file used `"`.

src/stan-users-guide/simple_linear_regression/run.R
```diff
@@ -0,0 +1,32 @@
...
+fit <- model$sample(data = stan_data, num_chains = 4, output_dir = "output")
+print(paste("running stan executable: ",model$exe_file()))
+print(fit$summary())
+
+#=============runs efficient version of regression_naive.stan========
```
bob-carpenter, line 27: Don't include big block comment separators.  If you want to call this out, print it out to user and it'll do double duty.

src/stan-users-guide/simple_linear_regression/run.R
```diff
@@ -0,0 +1,32 @@
...
+y <- rnorm(N, alpha + beta * x, sigma)
+
+stan_data <- list(N = N, x = x, y = y)
+
+print(sprintf("generating parameters are: alpha=%.1f, beta=%.1f, sigma=%.1f, N=%d",
+              alpha,beta,sigma,N))
+
+model <- cmdstan_model('regression_naive.stan')
+fit <- model$sample(data = stan_data, num_chains = 4, output_dir = "output")
+print(paste("running stan executable: ",model$exe_file()))
```
bob-carpenter, line 24: Spaces after commas.  Always.

src/stan-users-guide/simple_linear_regression/run.R
```diff
@@ -0,0 +1,32 @@
...
+print(fit$summary())
+
+#=============runs efficient version of regression_naive.stan========
+
+model2 <- cmdstan_model('regression_efficient.stan')
+fit2 <- model2$sample(data = stan_data, num_chains = 4, output_dir = "output")
```
bob-carpenter, line 30: I'd stick with defaults if they're reasonable.

src/stan-users-guide/simple_linear_regression/run.py
```diff
@@ -0,0 +1,35 @@
+""" Example program that generates data, compiles and runs regression.stan"""
+from cmdstanpy import CmdStanModel
+import numpy
+
+
+# Run from command line: Python regression_1.py
```
bob-carpenter, line 6: Don't document language.

src/stan-users-guide/simple_linear_regression/run.py
```diff
@@ -0,0 +1,35 @@
+""" Example program that generates data, compiles and runs regression.stan"""
```
bob-carpenter, line 1: There's no regression.stan.  I'd suggest keeping titles more vague so they don't go stale.

src/stan-users-guide/simple_linear_regression/run.py
```diff
@@ -0,0 +1,35 @@
...
+
+output = "generating parameters are: alpha={:.1f}, beta={:.1f}, sigma={:.1f}, n={:d}" 
+
+print(output.format(alpha,beta,sigma,n))
+
+stan_data = {'N': n, 'x': x, 'y': y}
+
+stan_program = CmdStanModel(stan_file='regression_naive.stan')
+stan_program.compile()
+fit = stan_program.sample(data=stan_data, output_dir='output')
```
bob-carpenter, line 23: Is output the default?  If so, just use that.

src/stan-users-guide/simple_linear_regression/run.py
```diff
@@ -0,0 +1,35 @@
...
+print("running stan executable: ",stan_program.exe_file)
+print(fit.summary())
+
+#=============runs efficient version========
+
+stan_program2 = CmdStanModel(stan_file='regression_efficient.stan')
+stan_program2.compile()
+fit2 = stan_program2.sample(data=stan_data, output_dir='output')
+print("running stan executable: ",stan_program2.exe_file)
```
bob-carpenter, line 32: Spaces after commas.

bob-carpenter: After doing a couple of these, we're going to want to pull all this cut-and-paste repeated code up into something higher order.  But I think we need to see more complicated examples.

src/stan-users-guide/simple_linear_regression/run.R
```diff
@@ -0,0 +1,32 @@
...
+alpha <- 2
+beta <- 3
+sigma <- 1
+N <- 100
+x <- runif(N, 0, 10)
+y <- rnorm(N, alpha + beta * x, sigma)
+
+stan_data <- list(N = N, x = x, y = y)
+
+print(sprintf("generating parameters are: alpha=%.1f, beta=%.1f, sigma=%.1f, N=%d",
```
bob-carpenter, line 19: This is going to be a pain if we try to do this in general when the parameter sizes get big.  Also, it's a big maintenance effort with naming.  I'd just skip the printing of the simulated parameters by default.  A much better output in general (that someone like @jgabry could figure out) would be to line up simulated with posterior means/medians/intervals.

src/stan-users-guide/.gitignore
```diff
@@ -0,0 +1,44 @@
+#stan intermediate files
```
bob-carpenter, line 1: I'm pretty sure .gitignore should only go at the top of the project.

src/stan-users-guide/simple_linear_regression/output/.gitignore
breckbaldwin: output/.gitignore ignores .csv and .txt files that are generated by running cmdstanR and cmdstanPy. Having it higher would filter below that or I guess I could put paths in but that gets ugly too. 

In docs/.gitignore gets added

```
#output files for footnotes
src/stan-users-guide/*/output/*.txt
src/stan-users-guide/*/output/*.csv
```
If you are ok with that solution then I'll do it that way. 
I don't have a better idea.

src/stan-users-guide/simple_linear_regression/regression_efficient.stan
breckbaldwin: I didn't realize _hat designated a point estimate. I wanted to distinguish from the generating programs, can I call them alpha_truth or something else? I think it is confusing to have the same variable have very different meanings.

src/stan-users-guide/simple_linear_regression/regression_efficient.stan
breckbaldwin: I missed the *, but kept the parameter constraint because I thought it was standard.

The usage in the users-guide is inconsistent:

```
stan-users-guide$ grep '<lower=' *.Rmd | wc -l
     324
stan-users-guide$ grep '<lower =' *.Rmd | wc -l
      45

```
Spaces matches R guidelines. I'll change it as I go but I am concerned that Andrew won't like it.

src/stan-users-guide/simple_linear_regression/output/.gitignore
bob-carpenter: Cool---didn't know you could even put .gitignore lower down.  I don't think we do it anywhere else, but it makes sense here.  

Where are these models going to go?  Will they be under src/stan-users-guide or under something like src/stan or src/footnotes (I really really don't like "footnotes" as the name of the directory for example models---it makes it sound very secondary)?  The bibtex is on the outside.

src/stan-users-guide/simple_linear_regression/regression_efficient.stan
bob-carpenter: Overloading is the way statisticians other than probability theorists roll w.r.t. naming.  It's very confusing for beginners.  I used to complain about this to no end.

First, the variables in the model should absolutely just be called `alpha`, `beta`, etc.  Anyting else is going to be too confusing to users.  Then their posterior means would be called alpha-hat (the posterior mean is the standard Bayesian estimator of a parameter as it minimizes expected square error).

I'd also just call the simulated value `alpha` in the R program.  The acceptable alternative would be `alpha_sim` but it starts getting clunky fast when you start decorating uses of `alpha` (hence the overloaded notation in the first place).

bob-carpenter:
Also, the footnote files can't just be spread across the top-level directory.  They should go under `src/examples` or `src/footnotes` or something like that.

src/stan-users-guide/.gitignore
bob-carpenter: I see I was wrong about that.  So it's fine leaving them lower.  Thanks for adding.

src/stan-users-guide/simple_linear_regression/regression_efficient.stan
breckbaldwin: How about alpha_s on the data generation side? I suffered at the hands of this particular convention and I'd like to not perpetuate it.

src/stan-users-guide/simple_linear_regression/regression_efficient.stan
breckbaldwin: Don't I have to unit scale the data then?

breckbaldwin: Is there a stan linter? I started to write one just to catch this stuff but before I go any further I thought I'd ask.

src/stan-users-guide/simple_linear_regression/run.py
breckbaldwin: This is a problem, especially right now since the code is being gone over so many times. I'll try and keep things generic--good idea. This is why I went with run.R and run.py. I hesitate to make the stan programs generic but I'll defer. 

How about I do another footnote dir or two and see where we are at.

src/stan-users-guide/simple_linear_regression/regression_efficient.stan
bob-carpenter: Yes, the user's guide inconsisent.  I think it should have spaces everywhere.  I think @andrewgelman is OK with that, but we can ask.  He usually sees these pings.

src/stan-users-guide/simple_linear_regression/run.py
breckbaldwin: It puts the files in a tmp dir and deletes them one the python process ends. Not sure what the right thing to do is since the output files have different information than what goes to disk.

src/stan-users-guide/simple_linear_regression/regression_efficient.stan
bob-carpenter: Yes, the data should be generated from the prior.   There's no way to generate the predictors from the prior because there's no model there.  So the usual approach is to consider data sets with independent predictors or ones with strongly correated predictors (by drawing from a high-correlation multinormal).

src/stan-users-guide/simple_linear_regression/run.R
breckbaldwin: I can mirror the estimates with the generating params easily enough. I'll ask Jonah if he has better ideas.

breckbaldwin:
> Also, the footnote files can't just be spread across the top-level directory. They should go under `src/examples` or `src/footnotes` or something like that.

Agreed, now that we are not linked to chapters no sense crapping up that level. I'll shove them in `src/footnotes`

andrewgelman:
I'm not sure what the normal(5,10) is doing here.  But, yeah, we'll have to specify the x's somehow and also beta and sigma.  I guess we could follow Bob's suggestion and split the difference and simulate the x's from a unit multivariate normal with off-diagonal correlations of 0.5?  I dunno, I hadn't thought too hard about the x's.

breckbaldwin:
I threw in the priors in to just have something to move the ball forward. 

What ever you folks want for examples.

andrewgelman:
In an earlier post I suggested normal(0,1) priors for everything unless explictily decided otherwise.

src/stan-users-guide/simple_linear_regression/regression_efficient.stan
breckbaldwin: I could use review on the multi-predictor case for 
docs/src/stan-users-guide/example_code/simple_linear_regression/run.R line 93 onwards. 

I am not sure I am generating the simulated data correctly.

thanks

breckbaldwin: This is in the footnotes branch. I am trying to get it merged.

breckbaldwin: I am going to resolve this conversation just to try and get the changes merged. Happy to revisit.

breckbaldwin:
@bbbales2 @bob-carpenter 

I think I incorporated revisions pretty closely to what you all wanted. 

Trying to get the PR done so I can continue to work more incrementally. I'll get multinomial logistic regression in once this is done--it is quite a bit more complex and we should discuss when I get it in a branch. 

I rebased so it ought to be minimally complex. 

thanks
Breck

bbbales2:
The diff is showing 2347 changed files. I was scrolling through and a lot of them are from the 2_22 release. Can you pull the rendered html stuff out of this pull request? That way I can actually get a sense of what files have changed (and then review those).

breckbaldwin:
> The diff is showing 2347 changed files. I was scrolling through and a lot of them are from the 2_22 release. Can you pull the rendered html stuff out of this pull request? That way I can actually get a sense of what files have changed (and then review those).

Thanks for looking at it, how are you generating the diff? I am unclear on what I need to change/remove from ??the branch?? 

All my changes are in docs/src/stan-users-guide/example_code and docs/src/stan-users-guide/regression.rmd. 

B

bbbales2:
> Thanks for looking at it, how are you generating the diff?

At the top of this page. How I do reviews is go to the top, click "Files changed" and read through the changes making comments.

How many files have you modified? Since nobody else is working on this branch yet, you could just make a new branch in a clean repo, add your changes there (don't use `git commit -a` or you'll pick up extraneous stuff), and then open a new pull request (closing this one and referencing it once you verify the new one only shows the changes you made).

breckbaldwin:
> 
> How many files have you modified? Since nobody else is working on this branch yet, you could just make a new branch in a clean repo, add your changes there (don't use `git commit -a` or you'll pick up extraneous stuff), and then open a new pull request (closing this one and referencing it once you verify the new one only shows the changes you made).

I have modified one file, docs/src/stan-users-guide/regression.Rmd, added 31 files or so in docs/src/stan-users-guide/example_code/. docs/src/stan-users-guide/example_code/README.txt explains what is going on.

New PR with branch footnotes2 is at https://github.com/stan-dev/docs/pull/250

bbbales2:
Cool I'm just going to close this branch since the other one is simpler.

PR #201, opened 2020-06-12
Add pedantic mode doc to reference manual

rybern:
#### Submission Checklist

- [X] Builds locally 

#### Summary

I added the pedantic mode doc to the end of the reference manual (after Diagnostic Mode). Let me know what you think.

#### Copyright and Licensing

### DIFF
--- a/src/reference-manual/pedantic-mode.Rmd
+++ b/src/reference-manual/pedantic-mode.Rmd
@@ -1,10 +1,10 @@
 
 
-# Pedantic Mode
+# Pedantic mode
 
-Pedantic Mode is a compilation option built into Stanc3 that warns you about potential issues in your Stan program.
+Pedantic mode is a compilation option built into Stanc3 that warns you about potential issues in your Stan program.
 
-For example, if you compile the following program with Pedantic Mode:
+For example, consider the following program.
 
     data {
       int N;
@@ -18,7 +18,7 @@ For example, if you compile the following program with Pedantic Mode:
       x ~ normal(mu, sigma);
     }
 
-The compiler will print the following to stderr:
+When pedantic mode is turned on, the compiler will produce the following warnings.
 
     Warning:
       The parameter sigma has no priors.
@@ -28,7 +28,7 @@ The compiler will print the following to stderr:
       A normal distribution is given parameter sigma as a scale parameter
       (argument 2), but sigma was not constrained to be strictly positive.
 
-Here are the kinds of issues that Pedantic Mode will find (which are described in more detail in following sections):
+Here are the kinds of issues that pedantic mode will find (which are described in more detail in following sections):
 
 -   *Distribution usages issues.* Distribution arguments don't match the distribution specification, or some specific distribution is used in an inadvisable way.
 -   *Unused parameter.* A parameter is defined but doesn't contribute to target.
@@ -39,14 +39,14 @@ Here are the kinds of issues that Pedantic Mode will find (which are described i
 -   *Variable is used before assignment.* A variable is used before being assigned a value.
 -   *Strict or nonsensical parameter bounds.* A parameter is given questionable bounds.
 
-Some important limitations of Pedantic Mode are listed at the end of this chapter.
+Some important limitations of pedantic mode are listed at the end of this chapter.
 
 
 ## Distribution argument and variate constraint issues
 
 When an argument to a built-in distribution certainly does not match that distribution's specification in the [Stan Functions Reference](https://mc-stan.org/docs/functions-reference/index.html), a warning is thrown. This primarily checks if any distribution argument's bounds at declaration, compile-time value, or subtype at declaration (e.g. `simplex`) is incompatible with the domain of the distribution.
 
-For example:
+For example, consider the following program.
 
     parameters {
       real unb_p;
@@ -59,7 +59,7 @@ For example:
 
 The parameter of `poisson` should be strictly positive, but `unb_p` is not constrained to be positive.
 
-This program results in the following warning:
+Pedantic mode produces the following warning.
 
     Warning at 'ex-dist-args.stan', line 6, column 14 to column 19:
       A poisson distribution is given parameter unb_p as a rate parameter
@@ -76,7 +76,7 @@ Pedantic mode checks for some specific uses of distributions that may indicate a
 Any use of uniform distribution generates a warning, except when the variate parameter's declared `upper` and `lower` bounds exactly match the uniform distribution bounds.
 In general, assigning a parameter a uniform distribution can create non-differentiable boundary conditions and is not recommended.
 
-For example:
+For example, consider the following program.
 
     parameters {
       real a;
@@ -89,7 +89,7 @@ For example:
 
 `a` is assigned a uniform distribution that doesn't match its constraints.
 
-This program results in the following warning:
+Pedantic mode produces the following warning.
 
     Warning at 'uniform-warn.stan', line 6, column 2 to column 20:
       Parameter a is given a uniform distribution. The uniform distribution is
@@ -119,7 +119,7 @@ See <https://mc-stan.org/docs/functions-reference/lkj-correlation.html> for deta
 A warning is generated when a parameter is declared but does not have any effect on the program.
 This is determined by checking whether the value of the `target` variable depends in any way on each of the parameters.
 
-For example:
+For example, consider the following program.
 
     parameters {
       real a;
@@ -131,7 +131,7 @@ For example:
 
 `a` participates in the density function but `b` does not.
 
-This program results in the following warning:
+Pedantic mode produces the following warning.
 
     Warning:
       The parameter b was declared but was not used in the density calculation.
@@ -142,7 +142,7 @@ This program results in the following warning:
 When numbers with magnitude less than 0.1 or greater than 10 are used as arguments to a distribution, it indicates that some parameter is not scaled to unit value, so a warning is thrown.
 See <https://mc-stan.org/docs/stan-users-guide/standardizing-predictors-and-outputs.html> for a discussion of scaling parameters.
 
-For example:
+For example, consider the following program.
 
     parameters {
       real x;
@@ -155,7 +155,7 @@ For example:
 
 The constants `-100` and `100` suggest that `x` is not unit scaled.
 
-This program results in the following warning:
+Pedantic mode produces the following warning.
 
     Warning at 'constants-warn.stan', line 6, column 14 to column 17:
       Argument -100 suggests there may be parameters that are not unit scale;
@@ -171,7 +171,7 @@ Control flow statements, such as `if`, `for` and `while`, should not depend on p
 This is likely to introduce a discontinuity into the density function.
 Pedantic mode generates a warning when any branching condition may depend on a parameter value.
 
-For example:
+For example, consider the following program.
 
     parameters {
       real a;
@@ -198,7 +198,7 @@ For example:
 
 The `if` and `for` statements are control flow that depend (indirectly) on the value of the parameter `m`.
 
-This program results in the following warning:
+Pedantic mode produces the following warning.
 
     Warning at 'param-dep-cf-warn.stan', line 11, column 2 to line 16, column 3:
       A control flow statement depends on parameter(s): a.
@@ -213,7 +213,7 @@ This pattern is not inherently an issue, but it is unusual and may indicate a mi
 
 Pedantic mode only searches for repeated statements, it will not for example generate a warning when a ~ statement is executed repeatedly inside of a loop.
 
-For example:
+For example, consider the following program.
 
     data {
       real x;
@@ -229,7 +229,7 @@ For example:
       b ~ normal(1, 1);
     }
 
-This program results in the following warning:
+Pedantic mode produces the following warning.
 
     Warning at 'multi-twiddle.stan', line 9, column 2 to column 19:
       The parameter a is on the left-hand side of more than one twiddle
@@ -244,7 +244,7 @@ This analysis depends on a [*factor graph*](https://en.wikipedia.org/wiki/Factor
 
 One limitation of this approach is that the compiler cannot distinguish between *modeled* `data` variables and other convenient uses of `data` variables such as data sizes or hyperparameters. This warning assumes that all data variables (except for `int` variables) are modeled data, which may cause extra warnings.
 
-For example:
+For example, consider the following program.
 
     data {
       real x;
@@ -273,7 +273,7 @@ For example:
 
 One prior is found for `a` and for `b`, while `c` only has a factor that touches a `data` variable and `d` has multiple priors.
 
-This program results in the following warning:
+Pedantic mode produces the following warning.
 
     Warning:
       The parameter c has no priors.
@@ -285,7 +285,7 @@ This program results in the following warning:
 
 A warning is generated when any variable is used before it has been assigned a value.
 
-For example:
+For example, consider the following program.
 
     transformed data {
       real x;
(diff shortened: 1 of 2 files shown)


bbbales2:
@rybern is pedantic in develop now? If so I can review this ezpz.

src/reference-manual/pedantic-mode.Rmd
```diff
@@ -0,0 +1,166 @@
...
+This pattern is not inherently an issue, but it is unusual and may indicate a mistake.
+
+Pedantic mode only searches for repeated statements, it will not for example generate a warning when a ~ statement is executed repeatedly inside of a loop.
+
+
+## Parameters with zero or multiple priors
+
+A warning is generated when a parameter appears to have greater than or less than one prior distribution factor.
+
+This analysis depends on a [*factor graph*](https://en.wikipedia.org/wiki/Factor_graph) representation of a Stan program. A factor F that depends on a parameter P is called a *prior factor for P* if there is no path in the factor graph from F to any data variable except through P.
```
bbbales2, line 125: Is there a reason for the CS definition here?

I think if I was doing an extended description here I would start with how this is different than checking that a parameter is not on the left hand side of a twiddle statement twice.

src/reference-manual/pedantic-mode.Rmd
```diff
@@ -0,0 +1,166 @@
...
+
+
+## Variables used before assignment
+
+A warning is generated when any variable is used before it has been assigned a value.
+This warning is also available as a standalone option to Stanc3, with the flag: `--warn-uninitialized`.
```
bbbales2, line 131: No interface stuff in the manual (I think stanc3 arguments count)

src/reference-manual/pedantic-mode.Rmd
```diff
@@ -0,0 +1,166 @@
...
+
+
+## Strict or nonsensical parameter bounds
+
+Parameters that are have strict `upper` and `lower` bounds can cause unmanageably large gradients in a density function, and may only be justified in a few cased.
```
bbbales2, line 136: From the error message itself, I think "Except when there are logical or physical constraints, it is very unusual for you to be sure that a parameter will fall inside a specified range" is more important than the infinite gradients.

src/reference-manual/pedantic-mode.Rmd
```diff
@@ -0,0 +1,166 @@
...
+
+
+#### Data variables
+
+The declaration information for `data` variables is currently not considered, so using `data` as incompatible arguments to distributions may not generate the appropriate warnings.
+
+
+#### Control flow dependent on parameters in nested functions
+
+If a parameter is passed as an argument to a user-defined function within another user-defined function, and then some control flow depends on that argument, the appropriate warning will not be thrown.
```
bbbales2, line 165: I'm for warning about these things as much as possible.

rybern:
@bbbales2 Thanks! Yes Pedantic Mode is in stanc3 master. I could definitely add more examples, I can probably get to that tomorrow.

src/reference-manual/pedantic-mode.Rmd
rybern: Ditto, I just left it for a future improvement because the solution is a bit involved.

src/reference-manual/pedantic-mode.Rmd
rybern: Makes sense

src/reference-manual/pedantic-mode.Rmd
rybern: The CS description is because it's the only way I know how to define it precisely, and it might be different under some circumstances from intuition.

I could a snippet about how this fires differently from twiddles, yeah, that is definitely a FAQ for good reason.

src/reference-manual/pedantic-mode.Rmd
```diff
@@ -0,0 +1,166 @@
...
+-   *Distribution usages issues.* Distribution arguments don't match the distribution specification, or some specific distribution is used in an inadvisable way.
+-   *Unused parameter.* A parameter is defined but doesn't contribute to target.
+-   *Large or small constant in a distribution.* Very large or very small constants are used as distribution arguments.
+-   *Control flow depends on a parameter.* Branching control flow (like if/else) depends on a parameter value.
+-   *Parameter has multiple twiddles.* A parameter is on the left-hand side of multiple twiddles.
+-   *Parameter has zero or multiple priors.* A parameter has more than one prior distribution.
+-   *Variable is used before assignment.* A variable is used before being assigned a value.
+-   *Strict or nonsensical parameter bounds.* A parameter is given questionable bounds.
+
+Some important limitations of Pedantic Mode are listed at the end of this section.
```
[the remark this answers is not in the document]
rybern: Could interpret section as the thing under the `#` heading, which is everything - is chapter better? Link to the subsection?

src/reference-manual/pedantic-mode.Rmd
```diff
@@ -0,0 +1,166 @@
...
+
+## Unused parameters
+
+A warning is generated when a parameter is declared but does not have any effect on the program.
+This is determined by checking whether the value of the `target` variable depends in any way on each of the parameters.
+
+
+## Large or small constants in a distribution
+
+When numbers with magnitude less than 0.1 or greater than 10 are used as arguments to a distribution, it indicates that some parameter is not scaled to unit value, so a warning is thrown.
```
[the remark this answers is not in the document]
rybern: Yeah, probably, but I don't feel qualified to an opinion on this one.

bbbales2:
> If there's a time crunch I'd be happy for you to make all of your suggested changes.

Let's just finish the doc before release next Monday. No need to merge today.

src/reference-manual/pedantic-mode.Rmd
bbbales2: In that case, I think you can leave a CS-y description, but can you also add an example of what it implies?

Like "this weird bit of code will throw a warning"

src/reference-manual/pedantic-mode.Rmd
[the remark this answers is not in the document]
bbbales2: The way this gets rendered in html is this sentence is at the end of a page. Sure there's a next page, but it's broken up.

bbbales2:
Also should there be something in the cmdstan docs about this?

rybern:
@bbbales2 Implemented most suggestions, including adding examples. I didn't add the snipped about twiddles vs. priors but the examples clarify that at least a bit.

Yeah there should probably be something in cmdstan docs.

src/reference-manual/pedantic-mode.Rmd
```diff
@@ -0,0 +1,166 @@
...
+See <https://mc-stan.org/docs/stan-users-guide/standardizing-predictors-and-outputs.html> for a discussion of scaling parameters.
+
+
+## Control flow depends on a parameter
+
+Control flow statements, such as `if`, `for` and `while`, should not depend on the value of the parameters to determine their branching conditions.
```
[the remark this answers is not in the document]
bbbales2: "Control flow statements, such as if, for and while, should not depend on the value of the parameters to determine their branching conditions. This is likely to introduce discontinuity into the density function." ->

"Control flow statements, such as if, for and while, should not depend on **parameters or functions of parameters**. This is likely to introduce **a** discontinuity into the density function."

I guess I'm waffling on this so the first is optional but definitely add the 'a'.

src/reference-manual/pedantic-mode.Rmd
```diff
@@ -0,0 +1,345 @@
...
+      consider rescaling with a multiplier (see manual section 22.12).
+
+
+## Control flow depends on a parameter
+
+Control flow statements, such as `if`, `for` and `while`, should not depend on the value of the parameters to determine their branching conditions.
+This is likely to introduce discontinuity into the density function.
+Pedantic mode generates a warning when any branching condition may depend on a parameter value.
+
+For example, in the following program:
```
bbbales2, line 167: Fragment!

src/reference-manual/pedantic-mode.Rmd
```diff
@@ -0,0 +1,345 @@
...
+
+
+## Parameters with multiple twiddles
+
+A warning is generated when a parameter is found on the left-hand side of more than one ~ statements (or an equivalent `target +=` conditional density statement).
+This pattern is not inherently an issue, but it is unusual and may indicate a mistake.
+
+Pedantic mode only searches for repeated statements, it will not for example generate a warning when a ~ statement is executed repeatedly inside of a loop.
+
+For example, in the following program:
```
bbbales2, line 207: Fragment!

src/reference-manual/pedantic-mode.Rmd
```diff
@@ -0,0 +1,345 @@
...
+    parameters {
+      real a;
+      real b;
+      real c;
+      real d;
+    }
+    model
+    {
+      a ~ normal(0, 1); // this is a prior
+      x ~ normal(a, 1); // this is not a prior, since data is involved
```
bbbales2, line 250: Oooh, I see kinda what is going on.

This check makes me a bit nervous, but this isn't something for this release. Parameterizing priors in terms of data hyperpriors is super convenient.

Horseshoe priors do this, for instance.

Look on page 29 of [this](https://arxiv.org/pdf/1707.01694.pdf)

Find this code:
```
real < lower =0 > scale_icept ; # prior std for the intercept
real < lower =0 > scale_global ; # scale for the half -t prior for tau
real < lower =1 > nu_global ; # degrees of freedom for the half -t priors for tau
real < lower =1 > nu_local ; # degrees of freedom for the half - t priors for lambdas
real < lower =0 > slab_scale ; # slab scale for the regularized horseshoe
real < lower =0 > slab_df ; # slab degrees of freedom for the regularized horseshoe
```

All those parameters are data-defined knobs for priors.

src/reference-manual/pedantic-mode.Rmd
```diff
@@ -0,0 +1,345 @@
...
+      The parameter c has no priors.
+    Warning:
+      The parameter d has 2 priors.
+
+
+## Variables used before assignment
+
+A warning is generated when any variable is used before it has been assigned a value.
+
+For example, in the following program:
```
bbbales2, line 275: Fragment!

src/reference-manual/pedantic-mode.Rmd
```diff
@@ -0,0 +1,345 @@
...
+
+
+## Strict or nonsensical parameter bounds
+
+Except when there are logical or physical constraints, it is very unusual for you to be sure that a parameter will fall inside a specified range.
+A warning is generated for all parameters declared with the bounds `<lower=.., upper=..>` except for `<lower=0, upper=1>` or `<lower=-1, upper=1>`.
+
+In addition, a warning is generated when a parameter bound is found to have `lower >= upper`.
+
+For example, in the following program:
```
bbbales2, line 299: Fragment!

src/reference-manual/pedantic-mode.Rmd
```diff
@@ -0,0 +1,345 @@
...
+-   *Strict or nonsensical parameter bounds.* A parameter is given questionable bounds.
+
+Some important limitations of Pedantic Mode are listed at the end of this chapter.
+
+
+## Distribution argument and variate constraint issues
+
+When an argument to a built-in distribution certainly does not match that distribution's specification in the [Stan Functions Reference](https://mc-stan.org/docs/functions-reference/index.html), a warning is thrown. This primarily checks if any distribution argument's bounds at declaration, compile-time value, or subtype at declaration (e.g. `simplex`) is incompatible with the domain of the distribution.
+
+For example, in the following program:
```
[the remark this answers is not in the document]
rybern: Whoops, yeah, meant to go back an fix this after copy/pasting..

src/reference-manual/pedantic-mode.Rmd
rybern: Right. I was talking with Bob about this as well. The issue is that we don't distinguish syntactically between modeled and modeled "data". Hyperparameters are an issue, another issue is size variables like `int N`. I manually ignore single `int`s for that reason. It would be cool if we could distinguish those somehow.

In the mean time, it's a good idea to add some explanation here.

src/reference-manual/pedantic-mode.Rmd
```diff
@@ -0,0 +1,363 @@
+
+
+# Pedantic Mode
+
+Pedantic Mode is a compilation option built into Stanc3 that warns you about potential issues in your Stan program.
```
bob-carpenter, line 5: "Pedantic Mode" should not be capitalized.  More generally, technical terms and names of techniques should be lower case, like "no U-turn sampler" and "slice sampling" and so on.  Things like "Hamiltonian Monte Carlo" are not capitalized, they're just made up out of proper names.  Compare "Markov chain Monte Carlo" where the "chain" is lower case.

src/reference-manual/pedantic-mode.Rmd
```diff
@@ -0,0 +1,363 @@
...
+
+For example, if you compile the following program with Pedantic Mode:
```
bob-carpenter, line 7: As @bbbales2 says, these are sentence fragments.  Grammatically, a dislayed formula is a noun and a displayed equation is like an embedded clause (and we can use clauses like nouns in the way we can use statements like expressions in programming languages, so it gets complicated).   

The way to fix this is to use something like

```
Consider compiling the following program with pedantic mode.

blah-blah blah

This program provides the following output.

blah-blah-blah

File: src/reference-manual/pedantic-mode.Rmd [hunk 16]
```diff
@@ -0,0 +1,363 @@
...
+      int N;
+      real x[N];
+    }
+    parameters {
+      real sigma;
+    }
+    model {
+      real mu;
+      x ~ normal(mu, sigma);
+    }
```
bob-carpenter, line 19: This all needs to be escaped for markdown.  So do the blocks of program output.  

The rendered HTML and pdf should be a mess for this.

src/reference-manual/pedantic-mode.Rmd
```diff
@@ -0,0 +1,363 @@
...
+
+The compiler will print the following to stderr:
+
```
bob-carpenter, line 22: There should not be references to `stderr` as that's not quite how Stan works.  The services make callbacks to functions passed in by the interfaces, which can then do whatever they like with it.  We send errors to one callback and non-errors to another.  I think CmdStan sends errors to stderr, but I'm not even sure about that.   R and Python intercept stderr and stdout to deal with for their REPL environments.

src/reference-manual/pedantic-mode.Rmd
bbbales2: Oh my, markdown. I was checking this by rendering the html to avoid these issues, and it's rendered preformatted!

There must be an equivalence between indenting for preformatting and surrounding with ticks. Yeah we should surround this stuff with ticks.

bob-carpenter: D'oh!  I forgot that four-space indent worked like ticks!  I should've rendered it myself.  Presumably that also goes through to the LaTeX.  I'm pretty sure we've used ticks almost everywhere else.

rybern: Yeah, I usually indent, since Pandoc and org-markdown both produce indent. Sounds like I should use ticks for consistency?

src/reference-manual/pedantic-mode.Rmd
rybern: I went with:

> For example, consider the following program.
> 
> blah-blah blah
> 
> Pedantic mode produces the following warning.
> 
> blah-blah-blah

src/reference-manual/pedantic-mode.Rmd
bbbales2: Yeah go ahead and backtick. Saves me from having to learn.

src/reference-manual/pedantic-mode.Rmd
bbbales2: blah-blah-blah approved!

Issue #19, opened 2018-12-23
manual longer term


bob-carpenter:
This is the issue for things that would be nice to get into the manual in the future.  Rather than closing this issue, we'll just tick things off as part of the regular round of "next manual" issues.

PR #213, opened 2020-07-04
Initial attempt at documenting new ODE interface (design-doc #19)

bbbales2:
#### Submission Checklist

- [x] Builds locally 

#### Summary

Here's a rough attempt at rewriting the ODE docs for the new interfaces.

@wds15 add this to your list to review. The user's guide portion is pretty rough. I don't think we should plan on doing anything more than a compatibility rewrite before the next release, but I think we should rewrite this shortly.

Issues:
1. There's a missing plot currently. I have a pull req. to remake it here: https://github.com/stan-dev/docs/issues/160 , and but the plot appears to be as a repo as a .pdf but that's not rendering on the web ([here](https://github.com/stan-dev/docs/blob/master/docs/2_23/stan-users-guide/img/sho-ode-trajectory.pdf)). Can I just convert it to a .png?

2. We can get rid of the old interface stuff in the user's guide, and it doesn't take up much space in the language reference, however, I think we should keep the full docs around in the function's guide since even though this is a deprecated feature there are going to be lots of models out there using `integrate_ode_X` functions. How do we demarcate this in the function guide? Presumably we need to make it easy for people to find that page, but also easy once they've found that page to realize that they should be using the new interfaces (on a different page). 

#### Copyright and Licensing


bbbales2:
@mitzimorris adding you here. I don't think you need to review the text so much (let's let @wds15 take a pass or two first), but do you have suggestions on #1 and #2 above?

src/functions-reference/higher-order_functions.Rmd
```diff
@@ -141,6 +142,138 @@ Stan's automatic differentiation.
 
 ## Ordinary Differential Equation (ODE) Solvers {#functions-ode-solver}
 
+Stan provides built-in ordinary differential equation (ODE) solvers.
+Because the first argument to each of the solvers is a function itself,
+the ODE solvers in Stan are considered high order functions.
```
SteveBronder, line 147:
```suggestion
Stan provides several higher order functions for solving Ordinary Differential Equations (ODEs).
```

src/functions-reference/higher-order_functions.Rmd
```diff
@@ -141,6 +142,138 @@ Stan's automatic differentiation.
...
+
+### Specifying an Ordinary Differential Equation as a Function
+
+A system of ODEs is specified as an ordinary function in Stan within
+the functions block. The ODE system function must have the same return
+type and first two arguments as the following function:
```
SteveBronder, line 153:
```suggestion
In Stan, a system of ODEs is specified as a function. The ODE system function must have the a `vector` return type with the first two arguments to the function being a `real` and `vector` such as:
```

src/functions-reference/higher-order_functions.Rmd
```diff
@@ -141,6 +142,138 @@ Stan's automatic differentiation.
...
+vector must match the length of the state input into the function.
+
+The arguments to this function are:
+
+*   *`time`*, the time to evaluate the ODE system
+
+*   *`state`*, the state of the ODE system at the time specified
+
+*   *...*, all other arguments passed through from the integrate call. Each
+argument can be any type.
```
SteveBronder, line 173: Before this section I would show an example signature for the ODE integrators because

> all other arguments passed through from the integrate call

Doesn't really make sense without knowing what the integrators look like

```suggestion

Where the arguments to the function are:

*   *`time`*, the time to evaluate the ODE system

*   *`state`*, the state of the ODE system at the time specified

*   *...*, Any additional arguments needed in the ODE. Each of these can be arguments of any type.

The ODE system function should return the derivative of the `state` with
respect to `time` at the `time` and `state` provided. The length of the returned
vector must match the length of the state input into the function.
```

src/functions-reference/higher-order_functions.Rmd
```diff
@@ -141,6 +142,138 @@ Stan's automatic differentiation.
...
+
+### Non-Stiff Solver
+
+<!-- vector[]; ode_rk45; (function ode, vector initial_state, real initial_time, real[] times, ...); -->
+\index{{\tt \bfseries ode\_rk45 }!{\tt (function ode, real[] initial\_state, real initial\_time, real[] times, ...): vector[]}|hyperpage}
+
+`vector[]` **`ode_rk45`**`(function ode, vector initial_state, real initial_time, real[] times, ...)`<br>\newline
+Solves the ODE system for the times provided using the Runge Kutta
+Dopri algorithm with the implementation from Boost.
```
SteveBronder, line 182: Should there be a link to wiki or boost for this method? Or is the user expected to know this solver already?

src/functions-reference/higher-order_functions.Rmd
```diff
@@ -141,6 +142,138 @@ Stan's automatic differentiation.
...
+differentiation formula (BDF) method with the implementation from
+CVODES with additional control parameters for the CVODES solver.
+
+### Arguments to the ODE solvers
+
+The arguments to the ODE solvers in both the stiff and non-stiff cases
+are as follows.
+
+*   *`ode`*: function literal referring to a function specifying  the
+system of differential equations with signature:
```
SteveBronder, line 231: What does function literal mean? Should we use that same term when talking about the ODE function above?

SteveBronder: Is this section the same as the one above?

src/stan-users-guide/odes.Rmd
```diff
@@ -52,52 +51,57 @@ sequence of times $t_0 < t_1 < t_2 < \cdots$.
...
 ### Strict Signature {-}
 
-The function defining the system must have exactly these argument
-types and return type.  This may require passing in zero-length arrays
-for data or parameters if the system does not involve data or
-parameters.  A full example for the simple harmonic oscillator, which
-does not depend on any constant data variables, is provided in the
-simple harmonic oscillator trajectory plot. 
+Every ODE system function has the same first two arguments and the
+same return type, but the third argument onwards may change be anything.
```
SteveBronder, line 79:
```suggestion
Every ODE system function must have the first two arguments be a `real` and `vector` type, but ODE functions may have any number and type of arguments as the rest of the function signature.
```

src/stan-users-guide/odes.Rmd
```diff
@@ -52,52 +51,57 @@ sequence of times $t_0 < t_1 < t_2 < \cdots$.
...
+
+For instance, all of these are valid ODE signatures:
+
+```
+vector myode1(real t, vector y, real a0);
+vector myode2(real t, vector y, int[] a0, vector a1);
+vector myode3(real t, vector y, matrix a0, real[] a1, row_vector a2);
+```
+
+And these are not allowed:
```
SteveBronder, line 89:
```suggestion
But these are not allowed:
```

src/stan-users-guide/odes.Rmd
```diff
@@ -257,124 +258,95 @@ Note that when the system is linear, it is a good idea to try
...
 
-Stan provides statistical inference for unknown initial states and/or
-parameters.  The ODE solver will be used deterministically to produce
-predictions, much like the linear predictor does in a generalized
-linear model.  These states will then be observed with measurement error.
-
+With noisy observations of the solution it is possible in Stan to estimate
+both parameters and initial conditions. As an example, this is a harmonic
+oscillator example where both the system parameter `theta` and initial state
+`y0` are parameters:
```
SteveBronder, line 282:
```suggestion
When the solution's observations are noisy it is possible in Stan to estimate
both the parameters and initial conditions. The example below is a harmonic
oscillator where both the system parameter `theta` and initial state
`y0` are parameters:
```

src/stan-users-guide/odes.Rmd
```diff
@@ -257,124 +258,95 @@ Note that when the system is linear, it is a good idea to try
...
+Compared to the program for simulation,
+the program to estimate parameters uses the `ode_rk45` function
 in the model block rather than the generated quantities block.  There
-are Cauchy priors on the measurement error scales `sigma` and standard
+are half-normal priors on the measurement error scales `sigma` and standard
 normal priors on the components of parameter array `theta` and initial
-state parameter array `y0`.  The solutions to the ODE are then
-assigned to an array `y_hat`, which is then used as the location in
+state parameter array `y0`.  The solutions to the ODE are
+assigned to `y_hat`, which is then used as the location in
```
SteveBronder, line 324: Is this example in the example repo as well? If so may be good to update there as well

src/stan-users-guide/odes.Rmd
```diff
@@ -257,124 +258,95 @@ Note that when the system is linear, it is a good idea to try
...
 ```
 
 As with other regression-like models, it's easy to change the noise
-model to be robust (e.g., Student-t distributed), to be correlated in
-the state variables (e.g., with a multivariate normal distribution),
-or both (e.g., with a multivariate Student-t distribution).
+model to something with heavier tails (e.g., Student-t distributed),
+to be correlated in the state variables (e.g., with a multivariate
+normal distribution), or both (e.g., with a multivariate Student-t
+distribution).
```
SteveBronder, line 337:
```suggestion
model to something with heavier tails (e.g., Student-t distributed),
correlation in the state variables (e.g., with a multivariate
normal distribution), or both heavy tails and correlation in the state variables (e.g., with a multivariate Student-t
distribution).
```

src/functions-reference/higher-order_functions.Rmd
bbbales2: I made the naming a little clearer, but yeah this is a searchable standard algorithm.

src/functions-reference/higher-order_functions.Rmd
bbbales2: > Is this section the same as the one above?

I left a copy of the old integrator docs there. The function signature will be deprecated but I figure it should still have docs so people know how to convert old models. So there is duplicate information.

src/functions-reference/higher-order_functions.Rmd
bbbales2: I moved the ODE system function definition below the signatures and the arguments to the integrator functions.

It's still slightly confusing because the first argument to the integrator function is the ODE system function, but there's a little inline doc there that hopefully works.

bbbales2:
Thanks @SteveBronder , those were useful comments. I resolved most of the simpler things and left open the big things if you wanna check them.

wds15:
Beginning of the stan user guides for odes it says

"Stan provides a built-in mechanism for specifying and solving systems of ordinary differential equations (ODEs). Stan provides three different integrators, tuned for solving non-stiff systems and for stiff systems."

can you please change that to

"Stan provides a built-in mechanism for specifying and solving systems of ordinary differential equations (ODEs). Three different integrators, tuned for solving non-stiff systems and for stiff systems are available."

wds15:
Also here: https://github.com/stan-dev/docs/blob/f145344cb539c90a09f6f536c9f5ac08f315e92e/src/stan-users-guide/odes.Rmd#L19

It states that stiff solver are slower, but more robust. Well... that's not quite right. For a non-stiff system the stiff solver will be slower than the non-stiff one, but whenever a system is stiff, then the non-stiff solver will be much slower than the stiff solver. If you could clarify that, this would be great. So for non-stiff problems the non-stiff solvers shine and leave behind the stiff solvers by a large margin, but if the problems are stiff, then the reverse is true.

src/functions-reference/higher-order_functions.Rmd
```diff
@@ -141,6 +142,131 @@ Stan's automatic differentiation.
 
 ## Ordinary Differential Equation (ODE) Solvers {#functions-ode-solver}
 
+Stan provides several higher order functions for solving initial value problems specified as Ordinary Differential Equations (ODEs).
+
+Solving an initial value ODE means given a set of differential equations $y' = f(y, t)$ and initial conditions $y(t_0)$, solving for $y$ at a sequence of times $t_0 < t_1 \leq t_2, \cdots \leq t_n$. $f(y, t)$ is referred to here as the ODE system function.
```
wds15, line 147: Here we should make the parameters $\theta$ explicit in f and y, I think... or say that we suppress them on purpose.

src/functions-reference/higher-order_functions.Rmd
```diff
@@ -141,6 +142,131 @@ Stan's automatic differentiation.
...
+The arguments to the ODE solvers in both the stiff and non-stiff cases
+are as follows.
+
+*   *`ode`*: ODE system function with signature `(real, vector, ...):vector`
+
+The arguments represent (1) time, (2) system state, and a sequence of arguments of any type passed unmodified from the integration function call. The return value contains the derivatives with respect to time of the state.
+
+*   *`initial_state`*: initial state, type `vector`,
+
+*   *`initial_time`*: initial time, type `int` or `real`,
```
wds15, line 210: We allow "int" as initial time... really? Is this cast into a real internally?

src/functions-reference/higher-order_functions.Rmd
```diff
@@ -141,6 +142,131 @@ Stan's automatic differentiation.
...
+
+*   `data`   *`abs_tol`*: absolute tolerance for the ODE  solver, type `real`,
+data only, and
+
+*   `data`   *`max_num_steps`*: maximum number of steps to take between output
+times in the ODE solver, type `int`, data only.
+
+### Specifying the ODE System Function
+
+The ODE system function must have the a `vector` return type, and the first two arguments must be a `real` and `vector` in that order. These two arguments can be followed by any number of other arguments of any type:
```
wds15, line 230: "The ODE system function must have as return type a" ....

src/functions-reference/higher-order_functions.Rmd
```diff
@@ -141,6 +142,131 @@ Stan's automatic differentiation.
...
+respect to time at the time and state provided. The length of the returned
+vector must match the length of the state input into the function.
+
+The arguments to this function are:
+
+*   *`time`*, the time to evaluate the ODE system
+
+*   *`state`*, the state of the ODE system at the time specified
+
+*   *...*, all other arguments passed through from the integrate call. Each
```
wds15, line 249: *`...`*

src/functions-reference/higher-order_functions.Rmd
```diff
@@ -141,6 +142,131 @@ Stan's automatic differentiation.
...
+The sizes must match, and in particular, the following groups are of
+the same size:
+
+*   state variables passed into the system function, derivatives
+returned by the system function, initial state passed into the
+solver, and length of each vector in the output,
+
+*   solution times and number of vectors in the output,
+
+## Old integrate_ode_X ODE Interface {#old-functions-ode-solver}
```
wds15, line 268: Maybe call this section "Deprecated `integrate_ode_*` ODE Interface" ?

src/stan-users-guide/odes.Rmd
```diff
@@ -212,14 +127,12 @@ simulate noisy observations given below.
 
 ```
 functions {
-  real[] sho(real t,
-             real[] y,
-             real[] theta,
-             real[] x_r,
-             int[] x_i) {
-    real dydt[2];
+  vector sho(real t,
```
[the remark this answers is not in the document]
bob-carpenter: Yes.  You can use stan mode.  Whether the resulting syntax highlighting is "nice" is another matter.

src/functions-reference/higher-order_functions.Rmd
bob-carpenter: Any function that can take a `real` argument should be able to take an `int` argument by casting it to `real`.  This does not (yet) apply to `real[]` vs. `int[]` arrays, I'm afraid, but should hold for primitives.

src/functions-reference/higher-order_functions.Rmd
bob-carpenter: It's probably best to start an appendix with deprecated functions as we do in the reference manual.

bob-carpenter:
1.  Just convert .pdf to .png and it should work on web and in the pdf.  You have to update the .Rmd to link to the new file.

2.  The reference manual should document the non-deprecated features.  There's an appendix for deprecated features, which should include all deprecated features that are still active.  I then think we should do the same with the functions reference.

src/stan-users-guide/odes.Rmd
```diff
@@ -383,33 +262,66 @@ signature.  The only difference is in the call to the integrator for
...
+For the RK45 solver, the default values for relative and absolute tolerance are
+both $10^{-6}$ and the maximum number of steps between outputs is
+one million. For the BDF and Adams solvers, the relative and absolute
+tolerances are $10^{-10}$ and the maximum number of steps between outputs is
+one hundred million.
+
+### Discontinuous ODE System Function {-}
+
+If there are discontinuities in the ODE system function, it is best
+to integrate the ODE to those discontinuities, stop the solver, and
```
bob-carpenter, line 305: "to those" -> "between the"

src/stan-users-guide/odes.Rmd
```diff
@@ -383,33 +262,66 @@ signature.  The only difference is in the call to the integrator for
...
+
+```
+if (t < t_lag)
+  return 0;
+else
+  ... return non-zero value...;
+```
+
+In general it is better to integrate up to `t_lag` in one solve and
+then integrate from `t_lag` onwards in another.
```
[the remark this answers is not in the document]
bob-carpenter: Cool.  I was wondering about that.  Definitely include this.

src/stan-users-guide/odes.Rmd
```diff
@@ -429,16 +341,86 @@ maximum absolute error allowed in a solution.
...
+
+This Stan program simulates noisy measurements from a simple harmonic
+oscillator.  The system of linear differential equations is coded as a
+matrix. The system parameters `theta` and initial state `y0` are read
+in as data along observation times `ts`.  The generated quantities
+block is used to solve the ODE for the specified times and then add
+random measurement error, producing observations `y_hat`. Because the
+ODEs are linear, we can use the `matrix_exp` function to solve the
+system.
+
```
bob-carpenter, line 421: Is there a way to get a steady state solution to a linear system using the matrix exponential?

src/stan-users-guide/odes.Rmd
```diff
@@ -429,16 +341,86 @@ maximum absolute error allowed in a solution.
...
+#### Sensitivity Analysis {-}
+
+The tolerances should be set low enough that setting them lower does
```
bob-carpenter, line 424: This is a confusing title.  This is not sensitivity analysis of the ODE, but of the ODE solver.  Can we call this something like "computational methodology" or "debugging" or "advice on setting tolerances" or something like that?

src/stan-users-guide/odes.Rmd
bbbales2: I checked and the footnotes stuff hasn't made it this far in the manual yet, so there's no updating to be done there.

bob-carpenter: @SteveBronder was talking about the `example-models` repo.   

The first footnotes PR is still pending:  https://github.com/stan-dev/docs/pull/164

src/stan-users-guide/odes.Rmd
```diff
@@ -52,136 +51,50 @@ sequence of times $t_0 < t_1 < t_2 < \cdots$.
...
-  real dydt[2];
+vector sho(real t,        // time
+           vector y,      // state
+           real theta) {
+  vector[2] dydt;
   dydt[1] = y[2];
-  dydt[2] = -y[1] - theta[1] * y[2];
+  dydt[2] = -y[1] - theta * y[2];
   return dydt;
 }
```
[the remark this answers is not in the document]
bbbales2: I agree style-wise, but right now:

```
return [...]';
```

is a little slow. See https://github.com/stan-dev/stanc3/issues/578.

src/functions-reference/higher-order_functions.Rmd
bbbales2: This is one of those things that's for convenience as much as anything. Something like:

```
ode_bdf(f, y0, 0, ts, ...)
```

will cause an error if only real is supported because the 0 gets interpreted as an int instead of a double. int should promote to var or double, but I vaguely remember having this problem at one point anyway. Not sure the exact conditions.

src/reference-manual/expressions.Rmd
```diff
@@ -1117,6 +1117,7 @@ literals.*
 `algebra_solver` | `vector, vector` | `real[], int[]` | `vector`
 `integrate_1d`, | `real, real, real[]` | `real[], int[]` | `real`
 `integrate_ode_X`, | `real, real[], real[]` | `real[], int[]` | `real[]`
+`ode_X`, | `vector, real, real[]` | | `vector[]`
```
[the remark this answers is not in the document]
bbbales2: Oops, I had a typo here (forgot `...`s):

I tried breaking this into two tables.

src/stan-users-guide/odes.Rmd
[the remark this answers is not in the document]
bbbales2: We aren't using this elsewhere in the manual, so I'm going to leave it off here.

src/stan-users-guide/odes.Rmd
```diff
@@ -257,124 +166,94 @@ Note that when the system is linear, it is a good idea to try
...
-assigned to an array `y_hat`, which is then used as the location in
+state parameter array `y0`.  The solutions to the ODE are
+assigned to `y_hat`, which is then used as the location in
 the observation noise model as follows.
 
 ```
-y_hat = integrate_ode_rk45(sho, y0, t0, ts, theta, x_r, x_i);
+vector[2] y_hat[T] = ode_rk45(sho, y0, t0, ts, theta);
 for (t in 1:T)
   y[t] ~ normal(y_hat[t], sigma);
```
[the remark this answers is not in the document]
bbbales2: This should be efficient. The ODE solve only happens once outside the for loop.

The elements of `y` are vectors so the sampling statement is already vectorized here.

bbbales2:
@wds15 @bob-carpenter, just warning, but didn't finish or push my changes here yet, so it'll probably be confusing if you try to make sense of my responses here (because I haven't pushed code yet).

bbbales2:
@bob-carpenter is there a standard way we link from the user's guide to the functions reference?

src/stan-users-guide/odes.Rmd
bbbales2: This makes more sense in the tolerance section so I moved the comment there.

src/stan-users-guide/odes.Rmd
```diff
@@ -383,33 +262,66 @@ signature.  The only difference is in the call to the integrator for
...
-the default maximum number of steps is `1e6` ($10^6$).
+`relative_tolerance` and `absolute_tolerance` control the stepsize and
+`max_num_steps` specifies the maximum number of steps the solver will
+take between output time points before giving up.
+
+For the RK45 solver, the default values for relative and absolute tolerance are
+both $10^{-6}$ and the maximum number of steps between outputs is
+one million. For the BDF and Adams solvers, the relative and absolute
+tolerances are $10^{-10}$ and the maximum number of steps between outputs is
+one hundred million.
```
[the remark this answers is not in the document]
bbbales2: I just took this default from the old code. I think you're right. I made an issue with computational motivation to make this smaller for both solvers: https://github.com/stan-dev/math/issues/1969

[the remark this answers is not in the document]
bbbales2: (But I'm going to leave this as it is for now)

src/stan-users-guide/odes.Rmd
```diff
@@ -257,124 +166,94 @@ Note that when the system is linear, it is a good idea to try
...
 specifies the Runge-Kutta solver (for non-stiff systems).
 
 Here, the ODE solver is called in the generated quantities block to
-provide a $10 \times 2$ array of solutions `y_hat` to
-which measurement error is added using the normal pseudo-random number
-generating function `normal_rng`.  The number of rows in the
-solution array is the same as the size of `ts`, the requested
-solution times.
-
-### Data versus Parameters {-}
```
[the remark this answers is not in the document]
bbbales2: I removed this because the new interface doesn't have this limitation (it can take any combination of data/param arguments).

In the new interface you might accidentally over-promote a data thing by defining it in transformed parameters, but whether it is defined in transformed parameters or transformed data the solver will still work.

src/stan-users-guide/odes.Rmd
```diff
@@ -383,33 +262,66 @@ signature.  The only difference is in the call to the integrator for
...
-The three control arguments are relative tolerance, absolute
-tolerance, and maximum number of steps.   The default values for
-relative and absolute tolerance are both `1e-6` ($10^{-6}$), and
-the default maximum number of steps is `1e6` ($10^6$).
+`relative_tolerance` and `absolute_tolerance` control the stepsize and
+`max_num_steps` specifies the maximum number of steps the solver will
+take between output time points before giving up.
+
+For the RK45 solver, the default values for relative and absolute tolerance are
+both $10^{-6}$ and the maximum number of steps between outputs is
```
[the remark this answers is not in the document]
bbbales2: Lol, symmetry is fun. I'm leaving this out though.

src/stan-users-guide/odes.Rmd
bbbales2: I think you'd do that by saying `y' = 0` for steady state means `0 = A * y` and them solving for `y`, which I guess is zero if `A` is full rank? I'm not sure but I don't think you'd do the matrix exponential.

src/stan-users-guide/odes.Rmd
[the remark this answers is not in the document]
bbbales2: > then he should at least output a solution in the vicinity or at the location of the discontinuity a value from the ODE solver. This forces the ODE solver to actually "look" at the time point in question as to what happens

I thought the ODE solvers interpolate to get their output points (and so the actual steps the solvers take are separate). I don't think adding output points will improve the situation.

bbbales2:
Alright I made my changes so this is ready for review again. The comments were super helpful btw.

@bob-carpenter if you want to leave the full final review to @wds15 that's fine by me, but I'd like you to take a look at the changes to the language reference and how I organized the function-reference appendix.

bob-carpenter:
There's not a good way to link to the functions reference.  It's doubly hard when they're getting published concurrently.  You can link to a specific version or to the top-level versionless one.  Obviously the versioned one will be stable and the other one less so.

bob-carpenter:
I'll just approve to leave the final review for @wds15.  Let me know if there's something you want to look at.  This is less of a big deal than the code, because we don't introduce backward compatibility issues, etc.  We can just change it in the next release if we don't like this one.

SteveBronder:
Should we wait to merge this till the code goes in?

wds15:
Will read this once more on Monday.

src/stan-users-guide/odes.Rmd
```diff
@@ -383,62 +271,164 @@ signature.  The only difference is in the call to the integrator for
...
+if (t < t_lag)
+  return 0;
+else
+  ... return non-zero value...;
+```
+
+In general it is better to integrate up to `t_lag` in one solve and
+then integrate from `t_lag` onwards in another. Mathematically, the
+discontinuity can make the problem ill-defined and the numerical integrator
+may behave erratically around it.
```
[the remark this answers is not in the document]
bbbales2: > then it's strongly advisable to output some function values in the vicinity of the discontinuity

I don't think this will improve the behavior of the integrator around those points though. Intermediate outputs are just interpolated from the actual path the solver took.

I guess I could suggest this if there was something to do with the output. Like make plots or something. I just don't think there's an obvious way to diagnose if there are problems or not here.

src/stan-users-guide/odes.Rmd
```diff
@@ -17,364 +16,253 @@ for stiff systems.
   [@CohenHindmarsh:1996; @SerbanHindmarsh:2005]
 
 For a discussion of stiff ODE systems, see the [stiff ODE
-section](#stiff-ode.section).  In a nutshell, the stiff solvers are
-slower, but more robust; how much so depends on the system and the
-region of parameter space.  The function signatures for Stan's ODE
+section](#stiff-ode.section). The function signatures for Stan's ODE
```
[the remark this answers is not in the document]
bbbales2: > This is due to the attempt of Adams to do a higher order approximation of the function in order to make large steps and in turn avoid excessive calls to the ODE RHS (whereas RK45 just calls brutally the ODE RHS very often).

I think we should work on a separate workflow ODE document because we have no real guidance on picking between the solvers. I'd like to just leave this for the future.

src/stan-users-guide/odes.Rmd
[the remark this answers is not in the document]
wds15: No, it really does make a difference. I had experienced this myself when coding dosing events as approximate delta peaks. Only when I added observation time points around the dosing time, only then the integrator picked up the dosing event. The big disadvantage of the approach you propose to restart is that you must restart the integration process with the initials being parameters whereas if you keep integrating in a single go, then - for pharmacokinetic problems at least - you can use data as the initial. This will dramatically speedup the solve process as the sensitivity system is a lot smaller as compared to making initials varying.

src/stan-users-guide/odes.Rmd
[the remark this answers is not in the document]
wds15: I find the text as is confusing as there is no guidance at all to pick between the two non-stiff ones. I would prefer to tell people: Start with RK45 in any case; if stiffness is a concern go with bdf and whenever the problem is non-stiff and the ODE RHS is expensive to evaluate, then try Adams (after having tried RK45).

This would reflect my intuition and I think is a fair summary.  If you think we are not ready for this... ok.

PR #262, opened 2020-08-19
Added docs for compiler optimizations

rybern:
#### Submission Checklist

- [X] Builds locally 

#### Summary

This PR adds documentation for compiler optimizations to the reference manual.

I'm keeping this as a draft PR until we resolve the interface in https://github.com/stan-dev/stanc3/pull/549.

Questions (for I'm guessing @bob-carpenter or @mitzimorris or @bbbales2):
- I put the compiler optimizations section under 'Usage' after 'Reproducibility', is that appropriate?
- I referenced some command line options (like `--optimize-numerically-close`) in order to describe their behavior, but I know that the reference manual isn't for describing the interface. Do I need to rename that behavior in order to describe it?

#### Copyright and Licensing


mitzimorris:
> I put the compiler optimizations section under 'Usage' after 'Reproducibility', is that appropriate?
sure.

> I referenced some command line options (like --optimize-numerically-close) in order to describe their behavior, but I know that the reference manual isn't for describing the interface. 

the CmdStan User's Guide has been added to the online docset - https://mc-stan.org/docs/cmdstan-guide/index.html

reference chapter on the compiler is here: https://mc-stan.org/docs/cmdstan-guide/stanc.html
and feel free to edit and correct the source file here:  https://github.com/stan-dev/docs/blob/master/src/cmdstan-guide/stanc.Rmd

src/reference-manual/compiler-optimizations.Rmd
```diff
@@ -0,0 +1,504 @@
...
+### Lazy code motion
+
+<a id="org0c605fa"></a> Lazy code motion rearranges the statements and expressions in a program with the goals of:
+
+-   Avoiding computing expressions more than once, and
+-   Computing expressions as late as possible (to minimize the strain on the working memory set).
+
+To accomplish these goals, lazy code motion will perform optimizations such as:
+
+-   Moving a repeatedly calculated expression its own variable (also referred to as *common-subexpression elimination*)
```
bbbales2, line 350: > expression its own variable

"expression to its own variable"

I'd also say Assigning rather than Moving, but if Moving is the appropriate lingo stick with that.

src/reference-manual/compiler-optimizations.Rmd
```diff
@@ -0,0 +1,504 @@
...
+-   Moving an expression outside of a loop, if it doesn't need to be in the loop (also referred to as *loop-invariant code motion*)
```
bbbales2, line 351: > loop, if it

"loop if it"

src/reference-manual/compiler-optimizations.Rmd
```diff
@@ -0,0 +1,504 @@
...
+
+  for (i in 1:10) {
+    x = sqrt(10);
+    y = sqrt(i);
+  }
+  z = sqrt(10);
+}
+```
+
+Compiler representation of program \*before lazy code motion (simplified from the output of `--debug-transformed-mir-pretty`):
```
bbbales2, line 371: Why `\*before`?

src/reference-manual/compiler-optimizations.Rmd
```diff
@@ -0,0 +1,504 @@
...
+  real z;
+  for(i in 1:10) {
+    x = sqrt(10);
+    y = sqrt(i);
+  }
+  z = sqrt(10);
+}
+```
+
+Compiler representation of program \*after lazy code motion (simplified from the output of `--debug-optimized-mir-pretty`):
```
bbbales2, line 386: `\*`

src/reference-manual/compiler-optimizations.Rmd
SteveBronder: I think moving is the right word here. If we said assignment I think we would have to say something like `Assign a repeatedly calculated expression a temporary'

src/reference-manual/compiler-optimizations.Rmd
```diff
@@ -0,0 +1,487 @@
+The Stanc3 compiler can attempt to optimize a Stan program as it is compiled. The optimized program has the same behavior as the unoptimized program, but it may be faster, more memory efficient, or more numerically stable.
```
rok-cesnovar, line 1:
```suggestion
The Stanc3 compiler can optimize the code of Stan model during compilation. The optimized model code behaves the same as unoptimized code, but it may be faster, more memory efficient, or more numerically stable.
```

src/reference-manual/compiler-optimizations.Rmd
```diff
@@ -0,0 +1,487 @@
...
+
+This section introduces the available optimization options and describes their effect.
+
+You can see a printout of a representation of the Stan program after the optimizations have been applied with the Stanc3 command-line option `--debug-optimized-mir-pretty`. You can see an analogous representation of the program before optimizations have been applied with `--debug-transformed-mir-pretty`.
```
rok-cesnovar, line 5:
```suggestion
You can see a pretty-printed representation of the Stan program after optimizations with the Stanc3 command-line option `--debug-optimized-mir-pretty`. To print an analogous representation of the program before optimizations use the `--debug-transformed-mir-pretty` flag.
```

rok-cesnovar: I would maybe move this below to not confuse the reader. I doubt many will try this, but I bet many will be interested in the details of optimizations.

src/reference-manual/compiler-optimizations.Rmd
```diff
@@ -0,0 +1,487 @@
...
+
+
+# Optimization levels
+
+To turn on optimizations, the user specifies the desired optimization *level*. The level specifies the set of optimizations which are turned on. Optimizations which are turned on are then applied in a specific order, and some are applied repeatedly.
```
rok-cesnovar, line 10:
```suggestion
To turn optimizations on, the user specifies the desired optimization *level*. The level specifies the set of optimizations to use. The chosen optimizations are used in a specific order, with some of them applied repeatedly.
```

src/reference-manual/compiler-optimizations.Rmd
```diff
@@ -0,0 +1,487 @@
...
+
+Optimization levels are specified by the numbers 0-1 and the 'experimental' tag:
+
+-   **O0** No optimizations are applied.
```
rok-cesnovar, line 14: Specify somehow that this is the default.

src/reference-manual/compiler-optimizations.Rmd
```diff
@@ -0,0 +1,487 @@
...
+-   **O1** Only optimizations which are simple, do not dramatically change the program, and are unlikely noticeably slow down compile times are applied.
```
rok-cesnovar, line 15:
```suggestion
-   **O1** Optimizations that are simple, do not dramatically change the program, and are unlikely to noticeably slow down compile times are applied.
```

src/reference-manual/compiler-optimizations.Rmd
```diff
@@ -0,0 +1,487 @@
...
+
+## Optimization descriptions
+
+
+### **O1** Optimizations
+
+
+#### Dead code elimination
+
+Dead code is code that does not have any effect on the behavior of the program. Code is not dead if it affects `target`, the value of any outside-observable variable like transformed parameters or generated quantities, or side effects such as print statements. Removing dead code can speed up a program by avoiding unnecessary computations.
```
rok-cesnovar, line 44:
```suggestion
Dead code is code that does not affect the behavior of the program. Code is not dead if it affects `target`, the value of any outside-observable variable like transformed parameters or generated quantities, or side effects such as print statements. Removing dead code can speed up a program by avoiding unnecessary computations.
```

src/reference-manual/compiler-optimizations.Rmd
```diff
@@ -0,0 +1,487 @@
...
+```
+log_prob {
+  int i;
+  FnPrint__("Hi!");
+}
+```
+
+#### Constant propagation
+
+Constant propagation replaces uses of a variable which is known to have a constant value `C` with that constant `C`. This removes the overhead of looking up the variable, and also makes many other optimizations possible (such as static loop unrolling and partial evaluation).
```
rok-cesnovar, line 88:
```suggestion
Constant propagation replaces the uses of a variable known to have a constant value `C` with that constant `C`. This optimization eliminates the overhead of looking up the variable and enables other optimizations (for example, static loop unrolling and partial evaluation).
```

src/reference-manual/compiler-optimizations.Rmd
```diff
@@ -0,0 +1,487 @@
...
+  data array[int, 100] a;
+  for(i in 1:100) {
+    a[i] = i;
+  }
+}
+```
+
+#### Copy propagation
+
+Copy propagation is similar to [expression propagation](#org4e2cadd), but only propagates variables rather than arbitrary expressions. This can reduce the complexity of the code for other optimizations such as expression propagation.
```
rok-cesnovar, line 128:
```suggestion
Copy propagation is similar to [expression propagation](#org4e2cadd) but only propagates variables rather than arbitrary expressions. This optimization can reduce the complexity of the code for other optimizations such as expression propagation.
```

src/reference-manual/compiler-optimizations.Rmd
```diff
@@ -0,0 +1,487 @@
...
+log_prob {
+  int i = 1;
+  int j = i;
+  int k = (i + i);
+}
+```
+
+#### Auto-differentiation level optimization
+
+Stan variables can have two auto-differentiation (AD) *levels*: AD or non-AD. AD variables carry gradient information with them, which allows Stan to calculate the log-density gradient, but they also have more overhead than non-AD variables. It is therefore inefficient for a variable to be AD unnecessarily. AD-level optimization sets every variable to be a floating point type unless its gradient is necessary.
```
rok-cesnovar, line 162:
```suggestion
Stan variables can have two auto-differentiation (AD) *levels*: AD or non-AD. AD variables carry gradient information with them, allowing Stan to calculate the log-density gradient, but they also have more overhead than non-AD variables. It is therefore inefficient for a variable to be AD unnecessarily. Therefore, AD-level optimization sets every variable as a floating point type unless its gradient is necessary.
```

src/reference-manual/compiler-optimizations.Rmd
```diff
@@ -0,0 +1,487 @@
...
+}
+
+log_prob {
+  data real x = (y + 1);
+}
+```
+
+#### Partial evaluation
+
+Partial evaluation searches for expressions that can be replaced with a faster, simpler, more memory efficient, or more numerically stable expression that has the same meaning.
```
rok-cesnovar, line 201:
```suggestion
Partial evaluation searches for expressions that we can replace with a faster, simpler, more memory efficient, or more numerically stable expression with the same meaning.
```

src/reference-manual/compiler-optimizations.Rmd
```diff
@@ -0,0 +1,487 @@
...
+```
+
+
+
+### **Oexperimental** Optimizations
+
+
+#### One step loop unrolling
+
+One step loop unrolling is similar to [static loop unrolling](#orge4599d2), but it only 'unrolls' the first iteration of a loop, and can therefore work even when the total number of iterations is not predictable. This can speed up a program by providing more opportunities for further optimizations such as partial evaluation and lazy code motion.
```
rok-cesnovar, line 240:
```suggestion
One step loop unrolling is similar to [static loop unrolling](#orge4599d2). However, it only 'unrolls' the first loop iteration and can therefore work even when the total number of iterations is not predictable. This optimization can speed up a program by providing further optimizations such as partial evaluation and lazy code motion.
```

src/reference-manual/compiler-optimizations.Rmd
```diff
@@ -0,0 +1,487 @@
...
+    for(i in (1 + 1):n) {
+      x = (x + i);
+    }
+  }
+}
+```
+
+#### Expression propagation
+
+<a id="org4e2cadd"></a> Constant propagation replaces uses of a variable which is known to have a constant value `E` with that constant `E`. This often results in recalculation of the expression, but provides more opportunities for further optimizations such as partial evaluation. Expression propagation is always followed by [lazy code motion](#org0c605fa) to avoid unnecessarily recomputing expressions.
```
rok-cesnovar, line 285:
```suggestion
<a id="org4e2cadd"></a> Constant propagation replaces the uses of a variable known to have a constant value `E` with that constant `E`. This often results in recalculating the expression but provides more opportunities for further optimizations such as partial evaluation. Expression propagation is always followed by [lazy code motion](#org0c605fa) to avoid unnecessarily recomputing expressions.
```

src/reference-manual/compiler-optimizations.Rmd
```diff
@@ -0,0 +1,487 @@
...
+    x = lcm_sym4__;
+    y = sqrt(i);
+  }
+  z = lcm_sym4__;
+}
+```
+
+#### Function inlining
+
+Function inlining replaces each function call to each user-defined function `f` with the body of `f`. It does this by copying the function body to the call site and doing the appropriate renaming of the argument variables. This can speed up a program by avoiding the overhead of a function call and providing more opportunities for further optimizations (such as partial evaluation).
```
rok-cesnovar, line 393:
```suggestion
Function inlining replaces each function call to each user-defined function `f` with the body of `f`. It does this by copying the function body to the call site and appropriately renaming the argument variables. This optimization can speed up a program by avoiding the overhead of a function call and providing more opportunities for further optimizations (such as partial evaluation).
```

src/reference-manual/compiler-optimizations.Rmd
```diff
@@ -0,0 +1,487 @@
...
+  b = inline_sym1__;
+}
+```
+
+In this code, the `for` loop and `break` is used to simulate the behavior of a `return` statement. The value to be returned is held in `inline_sym1__`. The flag variable `inline_sym3__` indicates whether a return has occurred and is necessary to handle `return` statements nested inside loops within the function body.
+
+
+#### Static loop unrolling
+
+<a id="orge4599d2"></a> Static loop unrolling takes a loop that has a predictable number of iterations `X` and replaces it by writing out the loop body `X` times. The loop index in each repeat is replaced with the appropriate constant. This can speed up a program by avoiding the overhead of a loop and providing more opportunities for further optimizations (such as partial evaluation).
```
rok-cesnovar, line 452:
```suggestion
<a id="orge4599d2"></a> Static loop unrolling takes a loop with a predictable number of iterations `X` and replaces it by writing out the loop body `X` times. The loop index in each repeat is replaced with the appropriate constant. This optimization can speed up a program by avoiding the overhead of a loop and providing more opportunities for further optimizations (such as partial evaluation).
```

rok-cesnovar (commented):
I started reading and just went through it all. Just a few suggestions, overall looks good.

WardBrian:
This PR should link (and backlink from) https://mc-stan.org/docs/stan-users-guide/using-the-stan-compiler.html

WardBrian:
@SteveBronder have you taken over this PR? Ping me if you want help, since the optimization interface was merged I think it's important to get this doc in before 2.29

src/reference-manual/compiler-optimizations.Rmd
WardBrian, line 1: I think this whole page now makes more sense as a sub-section of src/stan-users-guide/using-stanc.Rmd, which didn't exist when the PR was first opened

WardBrian:
Incorporated into #474

PR #333, opened 2021-02-03
[WIP] Attempt to update ODE Section

bbbales2:
#### Submission Checklist

- [x] Builds locally 
- [x] Declare copyright holder and open-source license: see below

#### Summary

Here's an attempt to update the ODE section. There's all sorts of stuff to talk about. My attempt was to reframe the introduction in a way that would make it easier to do that.

#### Copyright and Licensing

Please list the copyright holder for the work you are submitting (this will be you or your assignee, such as a university or company): Columbia University

By submitting this pull request, the copyright holder is agreeing to license the submitted work under the following licenses:
- Code: BSD 3-clause (https://opensource.org/licenses/BSD-3-Clause)
- Documentation: CC-BY 4.0 (https://creativecommons.org/licenses/by/4.0/)

PR #334, opened 2021-02-03
ckrk integrator doc

yizhang-yiz:
#### Submission Checklist

- [x] Builds locally 

#### Summary
This PR partially answers @bbbales2 's call in https://github.com/stan-dev/docs/pull/333 to improve ODE documentation. It provides a short description of `ckrk` integrator (https://github.com/stan-dev/math/pull/2336) and its intended usage.

#### Copyright and Licensing

Please list the copyright holder for the work you are submitting (this will be you or your assignee, such as a university or company): Metrum Research Group


bbbales2 (approved):
Good

bbbales2:
Bib conflict. I'll get that real quick.

The abstract on this paper is informative. I like how it explicitly differentiates itself from rk45.

I think we still need some more specific introductory text. I will open another pull with an attempt at that.

bbbales2:
Oh woops, Cash karp isn't in lol. Well I'll add my edits here.

bbbales2:
@yizhang-yiz I added some text. What do you think?

Also, reading the abstract, ckrk is an rk45 method. Can we just replace our rk45 with ckrk?

src/stan-users-guide/odes.Rmd
```diff
@@ -1,11 +1,41 @@
...
-Stan provides a built-in mechanism for specifying and solving systems
-of ordinary differential equations (ODEs). Three different solvers,
-tuned for solving non-stiff systems and for stiff systems are available.
+Stan provides four different methods for solving systems of ordinary
+differential equations (ODEs). All of these methods will adaptively
+refine their solutions in order to satisfy the given tolerances.
+
+The upside to this is that as long as reasonable tolerances are provided
+and an ODE solver well-suited to the problem is chosen, the choice of timestep
+can be abstracted away. The downside is that it is not always clear from the
```
yizhang-yiz, line 9: if there's maxstep control, timestep will not be completely "abstracted away".

src/stan-users-guide/odes.Rmd
```diff
@@ -1,11 +1,41 @@
...
+outset what reasonable tolerances are, or which ODE solver is best suited
```
yizhang-yiz, line 10: choosing "reasonable tolerance" is a universal consideration, not just to adaptive step + tolerance: using fixed stepsize still leads to Q of "how accuracy I want my answer to be?". So I wouldn't say it's a downside of adaptive step & tolerance control.

src/stan-users-guide/odes.Rmd
```diff
@@ -1,11 +1,41 @@
...
+to each problem. Additionally, in the course of working with an ODE model in
+Stan the correct answer may change (for instance, the ODE may change).
```
yizhang-yiz, line 12: This is a non-issue. If model changes, methodology should change accordingly, as in the other parts of Stan model.

yizhang-yiz:
> Can we just replace our rk45 with ckrk?

rk45 is FSAL but ckrk is not. So when RHS is relatively expensive to solve rk45 should have an advantage.

src/stan-users-guide/odes.Rmd
bbbales2: > methodology should change accordingly

Yeah, and this is trying to make it clear that if you're using ODEs changing the solver is part of the methodology. It's entirely possible someone might not think of their ode solver as a critical choice (or know a methodology to pick one).

src/stan-users-guide/odes.Rmd
yizhang-yiz: We can instead focus on emphasizing the importance of tolerance, such that user should first think really hard and make decision on the accuracy of the solution _before_ even choosing integrators. The accuracy is tied to their problem, data, as well as expectation. Then they want to investigate the tolerance will give the accuracy they want. This thought process doesn't have to involve we say anything about adaptive stepping.

bbbales2:
Wikipedia says:

For [Dormand Prince](https://en.wikipedia.org/wiki/Dormand%E2%80%93Prince_method), "it uses only six function evaluations per step because it has the FSAL"

For [Cash Karp](https://en.wikipedia.org/wiki/Cash%E2%80%93Karp_method), "it uses six function evaluations to calculate fourth- and fifth-order accurate solutions"

So maybe the difference is: "Dormand and Prince chose the coefficients of their method to minimize the error of the fifth-order solution. This is the main difference with the Fehlberg method, which was constructed so that the fourth-order solution has a small error."

The difference in minimizing fourth order and fifth order error doesn't seem like a whole lot (it's not like with Adams where we get possibly an extra 5 orders or whatever), so since ckrk behaves better with stiffness I kinda think it should be the default (more likely to behave reasonably on more problems).

yizhang-yiz:
But it may also try too hard on systems that user know a priori shouldn't be stiff. rk45 may just give up throwing out max-step error.

yizhang-yiz:
@bbbales2 an additional operational argument against replacing rk45 with ckrk is that we haven't see much ckrk in action.

bbbales2:
> it may also try too hard on systems that user know a priori shouldn't be stiff. rk45 may just give up

I don't think either rk45 or ckrk should be giving up on non-stiff systems.

> we haven't see much ckrk in action

True, but it's the same family of integrators meant to handle quickly changing right hand sides better (or I forget the exact wording, but something like this).

So the reason I got to thinking about just replacing our current rk45 integrator again is writing the sentence:

```
The difference between `ckrk` and `rk45` is that `ckrk` should perform
better for systems that exhibit rapidly varying solutions.
```

Like that seems like a real fine hair to split. So why not ckrk if it works better with the funky stuff! If people have non-stiff systems with smooth solutions, that's Adams territory anyway.

yizhang-yiz:
> Like that seems like a real fine hair to split. So why not ckrk if it works better with the funky stuff! 

After dealing with somethings on my hand I can probably come up a model where rk45 works better than ckrk.

> If people have non-stiff systems with smooth solutions, that's Adams territory anyway.

Well keep in mind Adams doesn't have advantage when requested precision is low (actually there's another place Adams works better then rk45 but we're not seeing those applications in Stan).

src/stan-users-guide/odes.Rmd
```diff
@@ -4,30 +4,35 @@ Stan provides four different methods for solving systems of ordinary
...
 
-The upside to this is that as long as reasonable tolerances are provided
-and an ODE solver well-suited to the problem is chosen, the choice of timestep
-can be abstracted away. The downside is that it is not always clear from the
-outset what reasonable tolerances are, or which ODE solver is best suited
-to each problem. Additionally, in the course of working with an ODE model in
-Stan the correct answer may change (for instance, the ODE may change).
+Each solver has a simple interface with default tolerances and then an advanced
+one that allows the tolerances to be modified. If components of the solution
+need to be solved with high precision or if the solution goes very near zero,
```
yizhang-yiz, line 9: We may want to adjust wording to avoid confusion: if a low(like 1.e-4) solution is expected one can also choose `_tol` solvers.

bbbales2: I don't want to encourage the coarse tolerance -> high performance thing and I'm also sketched out about low precision gradients for HMC.

bbbales2: (So I don't want to encourage lower than default tolerances)

bbbales2:
Is there anything more descriptive than:

```
The difference between `ckrk` and `rk45` is that `ckrk` should perform
  better for systems that exhibit rapidly varying solutions.
```

that we can say to differentiate it from rk45? We really don't have a word count here.

src/stan-users-guide/odes.Rmd
yizhang-yiz: we doubt we have a lot of evidence to back the default tolerances. In many occasions they're simply not reasonable. In lotka-volterra model the tol is set to 1.e-3: how would one justify measuring the number of hares up to 1.e-6 precision?

bbbales2: Mm, fair, I changed it to:

```
If components of the solution need to be solved with higher precision, the
solution can be solved with lower than default tolerances, or if the solution
goes very near zero, then the advanced interface may be useful.
```

bbbales2:
@yizhang-yiz can you add function reference docs here? Also check if the ode solvers are mentioned by name in the reference manual anywhere too (and add these if necessary)

bbbales2:
@yizhang-yiz I made a couple changes (diff [here](https://github.com/stan-dev/docs/pull/334/commits/2fb7c3080b7fa20b5ceafa12d0a4ec24bb37f4f5)). If you're happy with this go ahead and merge

yizhang-yiz:
@bbbales2 changed some wording. I see now you've moved to the general ODE section, not just ckrk part.

Issue #383, opened 2021-08-24
revise variable typing in all doc


bob-carpenter:
#### Summary:

Include table describing variable type in user's guide to manual and upgrade all doc to include `offset` and `multiplier` and use new array syntax.

#### Description:

* [x] Describe variable types in reference manual.  Copy [table from user's guide](https://mc-stan.org/docs/2_27/stan-users-guide/basic-functions-section.html#type-declarations-for-functions) to reference manual.
    * block variable: constrained, sized
    * local variable: unconstrained, sized
    * function argument: unconstrained, unsized

* [x] Update all doc to use `array[N] T x` in place of `T x[N]`
* [x] include deprecation notice for `T x[N]` in reference manual.
* [x] update Stan program style guide in user's guide

#### Current Version:
v2.26.0

WardBrian:
First step of this is making sure the style guide actually represents what we want it to, I'm happy to then comb through and make sure everything looks good

PR #385, opened 2021-08-25
Revise style guide

WardBrian:
#### Submission Checklist

- [x] Builds locally 

#### Summary

This PR makes several changes to the style guide. The goal is to align more closely with how the language is formatted by `--auto-format` in stanc3 and in general be more declarative. There are fewer "if you prefer, do this, else, do this" statements -- users can happily ignore this guide anyway, so it seems best to present a single, consistent style while knowing full well that most people won't follow it 100%.

Following this PR I will do a pass through all our doc to make sure this is actually followed, and that deprecated features aren't used in examples (#383)

#### Copyright and Licensing

Simons Foundation

### DIFF
--- a/src/stan-users-guide/style-guide.Rmd
+++ b/src/stan-users-guide/style-guide.Rmd
@@ -6,7 +6,9 @@ recommendations for laying out programs in a text editor.  Although
 these recommendations may seem arbitrary, they are similar to those of
 many teams for many programming languages.  Like rules for typesetting
 text, the goal is to achieve readability without wasting white space
-either vertically or horizontally.
+either vertically or horizontally. This is the style used in the Stan
+documentation, and should align with the auto-formatting ability of 
+`stanc3`.
 
 ## Choose a consistent style
 
@@ -29,7 +31,8 @@ sacrificed is laying out expressions on a single line.
 
 The recommended file extension for Stan model files is `.stan`.
 For Stan data dump files, the recommended extension is `.R`, or
-more informatively, `.data.R`.
+more informatively, `.data.R`. For JSON output, the recommended
+extension is `.json`.
 
 ## Variable naming
 
@@ -37,10 +40,11 @@ The recommended variable naming is to follow C/C++ naming
 conventions, in which variables are lowercase, with the underscore
 character (`_`) used as a separator.  Thus it is preferred to use
 `sigma_y`, rather than the run together `sigmay`, camel-case
-`sigmaY`, or capitalized camel-case `SigmaY`.  Even matrix
-variables should be lowercased.
+`sigmaY`, or capitalized camel-case `SigmaY`. An exception is often
+made for terms appearing in mathematical expressions with standard
+names, like `A` for a matrix.
 
-The exception to the lowercasing recommendation, which also follows
+Another exception to the lowercasing recommendation, which follows
 the C/C++ conventions, is for size constants, for which the
 recommended form is a single uppercase letter.  The reason for this is
 that it allows the loop variables to match.  So loops over the indices of
@@ -123,10 +127,10 @@ outside of both loops, and used a total of `N` times.
 
 ## Parentheses and brackets
 
-### Optional parentheses for single-statement blocks {-}
+### Braces for single-statement blocks {-}
 
-Single-statement blocks can be rendered in one of two ways.  The fully
-explicit bracketed way is as follows.
+Single-statement blocks can be rendered in several ways.
+The **recommended** fully explicit bracketed way is as follows.
 
 ```
 for (n in 1:N) {
@@ -148,15 +152,15 @@ in the following example.
 for (n in 1:N) y[n] ~ normal(mu,1);
 ```
 
-These can be much harder to read than the first example. Only use this
-style if the statement is  simple, as in this example.  Unless
+These can be much harder to read than the first example. Unless
 there are many similar cases, it's almost always clearer to put
 each sampling statement on its own line.
 
-Conditional and looping statements may also be written without brackets.
+Conditional and looping statements may also be written without 
+brackets.
 
-The use of for loops without brackets can be dangerous.  For instance,
-consider this program.
+The use of loops and conditionals without brackets can be dangerous.  
+For instance, consider this program.
 
 ```
 for (n in 1:N)
@@ -175,6 +179,7 @@ for (n in 1:N) {
 y[n] ~ normal(mu,1);
 ```
 
+Therefore, we recommend using braces in all instances for clarity.
 
 
 ### Parentheses in nested operator expressions {-}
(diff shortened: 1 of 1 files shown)


src/stan-users-guide/style-guide.Rmd
```diff
@@ -123,40 +127,40 @@ outside of both loops, and used a total of `N` times.
...
+but should be avoided.
 
 ```
 for (n in 1:N)
   y[n] ~ normal(mu,1);
 ```
 
-Single-statement blocks can also be written on a single line, as
-in the following example.
+Finally, single-statement blocks can also be written on a single
```
bob-carpenter, line 149: I think we should just eliminate this kind of judgement call and say it should always have braces and newlines.

src/stan-users-guide/style-guide.Rmd
```diff
@@ -194,12 +199,11 @@ the bracketed form `(y[n] > 0) || (x[n] != 0)`.
...
 Vertical space is valuable as it controls how much of a program you
-can see.  The preferred Stan style is as shown in the previous
-section, not as follows.
+can see.  The preferred Stan style is with the opening brace appearing
+at the end of a line.
 
 ```
-for (n in 1:N)
-{
+for (n in 1:N){
```
bob-carpenter, line 206: Need space between `)` and `{`!

src/stan-users-guide/style-guide.Rmd
```diff
@@ -214,33 +218,34 @@ transformed parameters {
...
+own line.
 
 
 ## Conditionals
 
-Stan supports the full C++-style conditional syntax,
-allowing real or integer values to act as conditions, as follows.
+While Stan supports the full C++-style conditional syntax,
+allowing real or integer values to act as conditions, this
+should be avoided.
```
bob-carpenter, line 231: "real values should be avoided".  "this" is ambiguous on its own.

src/stan-users-guide/style-guide.Rmd
```diff
@@ -214,33 +218,34 @@ transformed parameters {
...
+
+The condtional expression should never be a value that is not produced
+ as the result of a comparison or boolean operation. Therefore, use
```
bob-carpenter, line 234: Seems to be an extra line-initial space here.  I don't like the general guidance, though.  I think we should allow something like this:

```stan
if (bernoulli_rng(0.5)) {
   // do something
}
```

I think it's also totally OK to have this:

```stan
data {
  array[N]  int<lower = 0, upper = 1> foo;
  ...
model {
  for (n in 1:N) {
    if (foo[n]) {
      // do something
    }
  }
}
```

I think all we need to say is that rather than a real expression `e`, we should have `e != 0`.

src/stan-users-guide/style-guide.Rmd
```diff
@@ -314,74 +319,98 @@ layout.
...
+Use a space after `if`s.  For instance, use `if (x < y) {...`, not
+`if(x < y){ ...`.
 
-### No space For function calls {-}
+### No space for function calls {-}
 
-There is no space between a function name and the function it applies
-to.  For instance, use `normal(0,1)`, not `normal (0,1)`.
+There should not be space between a function name and the arguments 
+it applies to.  For instance, use `normal(0, 1)`, not `normal (0,1)`.
```
bob-carpenter, line 350: The for instance seems to be for the wrong thing.  The removed line 352 was th right counterexample for this sentence.

src/stan-users-guide/style-guide.Rmd
```diff
@@ -314,74 +319,98 @@ layout.
...
 
 There should be spaces around binary operators.  For instance, use
 `y[1] = x`, not `y[1]=x`, use `(x + y) * z` not
 `(x+y)*z`.
 
+Unary operators are written without a space, such as in `-x`, `!y`.
+
+### No spaces in type constraints {-}
+Another exception to the above rule is when the operator is the 
+assignment operator `=` used inside a type constraint, such as
```
bob-carpenter, line 362: "when the operator is the assignment operator" -> "when the assignment operator"

src/stan-users-guide/style-guide.Rmd
```diff
@@ -314,74 +319,98 @@ layout.
...
+
+```
+real<lower=0> x;
+```
+
+Here, the preferred style is to omit spaces, except for those following commas as in
```
bob-carpenter, line 368: We only want to omit spaces around the `=`.  So rather than

```
real<lower=a*x+b> alpha;
```

I think we'd want

```
real<lower=a * x + b> alpha;
```

I still like it with spaces around the equals sign, myself.

src/stan-users-guide/style-guide.Rmd
```diff
@@ -314,74 +319,98 @@ layout.
...
+```
+real<lower=0, upper=10> x;
+```
+
 ### Breaking expressions across lines {-}
 
-Sometimes expressions are too long to fit on a single line.  In that case, the recommended form is to break *before* an operator,^[This is the usual convention in both typesetting and other programming languages. Neither R nor BUGS allows breaks before an operator because they allow newlines to signal the end of an expression or statement.]  aligning the operator to indicate scoping.  For example, use the following form (though not the content; inverting matrices is almost always a bad idea).
+Sometimes expressions are too long to fit on a single line.  In that case,
+ the recommended form is to break *before* an
```
bob-carpenter, line 376: extra line initial space

bob-carpenter (changes requested):
Almost there.

src/stan-users-guide/style-guide.Rmd
WardBrian: Good catch. Your suggestion aligns with how the compiler would format it currently, I just hadn't thought of more complicated expressions in the constraint

Of the three obvious options here (
```stan
real<lower=a*x+b> alpha;
real<lower=a * x + b> alpha;
real<lower = a * x + b> alpha;
```
)
I honestly don't have a super strong preference, but I think two is a nice balance between readable and short, even if it is slightly inconsistent. Then again, it's not like we're gonna be policing this style guide

src/stan-users-guide/style-guide.Rmd
WardBrian: I don't think so. The only change I made was to add a space after the comma in the 'correct' example

src/stan-users-guide/style-guide.Rmd
bob-carpenter: I'm OK with (2) as the official style.  It's not going to come up very often.  When these things get complicated, I tend to write functions for the bounds, which would look OK jammed up, as in 

```stan
real<lower=f(a, x, b)> alpha;
```

src/stan-users-guide/style-guide.Rmd
bob-carpenter: Sorry---I just reversed the polarities.  350 is definitely the right example.

bob-carpenter (approved):
Looks good.  Thanks.

PR #391, opened 2021-08-30
Corrected range of arctan

kwichmann:
Range of atan is not -pi to pi, but -pi/2 to pi/2.

#### Submission Checklist

- [ ] Builds locally 
- [ ] Declare copyright holder and open-source license: see below

#### Copyright and Licensing

Please list the copyright holder for the work you are submitting (this will be you or your assignee, such as a university or company):

By submitting this pull request, the copyright holder is agreeing to license the submitted work under the following licenses:
- Code: BSD 3-clause (https://opensource.org/licenses/BSD-3-Clause)
- Documentation: CC-BY 4.0 (https://creativecommons.org/licenses/by/4.0/)

PR #393, opened 2021-08-31
update inv_phi

spinkney:
#### Submission Checklist

- [ ] Builds locally 

#### Summary
Updated the inv_Phi() docs for https://github.com/stan-dev/math/pull/2566 and corrected the name of the `real-valued_basic_functions.Rmd` file that was corrupted due to #391. 

#### Copyright and Licensing

Please list the copyright holder for the work you are submitting (this will be you or your assignee, such as a university or company): Comscore, Inc.

### DIFF
--- a/src/functions-reference/real-valued_basic_functions.Rmd
+++ b/src/functions-reference/real-valued_basic_functions.Rmd
@@ -1001,8 +1001,10 @@ standard normal cumulative distribution function of x
 \index{{\tt \bfseries inv\_phi }!{\tt (T x): R}|hyperpage}
 
 `R` **`inv_Phi`**`(T x)`<br>\newline
-standard normal inverse cumulative distribution function of p,
-otherwise known as the quantile function
+Return the standard normal inverse cumulative distribution function of p,
+otherwise known as the quantile function. The algorithm is detailed in [@Wichura:1988]
+The algorithm provides provides precise results up to about 16 digits for values between (1e-16, 0.999999999).
+Values smaller than 1e-16 are untested. Values above 0.999999999 result in increasingly larger errors.
 
 <!-- R; Phi_approx; (T x); -->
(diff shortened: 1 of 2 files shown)


src/functions-reference/real-valued_basic_functions.Rmd
```diff
@@ -1001,8 +1001,10 @@ standard normal cumulative distribution function of x
 \index{{\tt \bfseries inv\_phi }!{\tt (T x): R}|hyperpage}
 
 `R` **`inv_Phi`**`(T x)`<br>\newline
-standard normal inverse cumulative distribution function of p,
-otherwise known as the quantile function
+Return the standard normal inverse cumulative distribution function of p,
```
bob-carpenter, line 1004: There's no variable p here.  The return is a value, not a function.  I think the first line of this could be rewritten as

```
Return the value of inverse standard normal cdf $\Phi^{-1}$ at the specified quantile `x`.
```

Then rather than "values between (L, U)", I'd say "quantile arguments in (L, U)".

And then rather than "Values smaller than 1e-16", I'd say "Quantile arguments below 1e-16 are untested; quantiles above 0.999999999 result in increasingly large errors". 

This is OK, but I wouldn't be surprised if accuracy was still good for arguments below 1e-16.

bob-carpenter (changes requested):
Thanks much for updating the doc.  My other comment was about the oxygen doc, but this is an even better place to put it.  

Just a couple minor nitpicks and then we can merge. Mostly about typing in English (I used to be a linguist working on type theory, so apologies in advance for being so picky about this kind of thing).

spinkney:
Np, I'm loving the feedback. Thanks for all your reviews!

bob-carpenter (approved):
Thanks!

PR #400, opened 2021-09-14
Add conda install instructions to the cmdstan guide

WardBrian:
#### Submission Checklist

- [x] Builds locally 

#### Summary

Adds instructions for how to install cmdstan via the [conda-forge distribution](https://github.com/conda-forge/cmdstan-feedstock). I'm not sure if we want this in the main doc or not, but we have it as the recommended way for [cmdstanpy](https://cmdstanpy.readthedocs.io/en/v0.9.77/installation.html#conda-users-recommended)

#### Copyright and Licensing

Simons Foundation

### DIFF
--- a/src/cmdstan-guide/installation.Rmd
+++ b/src/cmdstan-guide/installation.Rmd
@@ -25,6 +25,11 @@ to activate the CmdStan makefile from anywhere. To enable either
 of these, ensure your environment is activated (via
 `conda activate cmdstan`, or whichever name you used above).
 
+Issues with the conda installation are handled separately.
+Please report conda-specific problems
+[here](https://github.com/conda-forge/cmdstan-feedstock/issues).
+
+
 ## Source installation
 
 To install CmdStan from source you need:
--- a/src/cmdstan-guide/installation.Rmd
+++ b/src/cmdstan-guide/installation.Rmd
@@ -8,10 +8,10 @@ will install a pre-built version of CmdStan along with the required
 dependencies (i.e. a C++ compiler, a version of Make, and required
 libraries) detailed below under [Source installation].
 
-One can create a new conda environment (named `stan`)
+One can create a new conda environment (named `stan-env`)
 and install CmdStan with the following command.
 ```
-conda create -n stan -c conda-forge cmdstan
+conda create -n stan-env -c conda-forge cmdstan
 ```
 or install it in an existing environment with
 ```
@@ -27,8 +27,20 @@ Installing software directly in `base` is not recommended by conda, but may be
 appropriate for some users who do not intend to use multiple environments]
 with `conda activate <env>`, where `<env>` is the name of
 the environment CmdStan was installed into. In the first command
-above, the environment was named `stan`, but any name may be used
-so long as it you are consistent.
+above, the environment was named `stan-env`, but any name may be used
+so long as you are consistent.
+
+You can check which version of CmdStan is installed in your current
+environment by running the following command.
+```
+conda list cmdstan
+```
+Finally, you can specify a different version for installation by adding `=VERSION`
+to the end of the above commands.^[You can view available versions with the
+command `conda search -c conda-forge cmdstan`] For example,
+```
+conda install -c conda-forge cmdstan=2.27.0
+```
 
 Please report conda-specific install problems directly to the
 conda-forge issue tracker,
--- a/src/cmdstan-guide/installation.Rmd
+++ b/src/cmdstan-guide/installation.Rmd
@@ -8,25 +8,30 @@ will install a pre-built version of CmdStan along with the required
 dependencies (i.e. a C++ compiler, a version of Make, and required
 libraries) detailed below under [Source installation].
 
-One can create a new conda environment and install CmdStan with
-the following command.
-
+One can create a new conda environment (named `stan`)
+and install CmdStan with the following command.
 ```
-conda create -n cmdstan -c conda-forge cmdstan
+conda create -n stan -c conda-forge cmdstan
 ```
 or install it in an existing environment with
 ```
 conda install -c conda-forge cmdstan
 ```
 
-With the conda installation, one can use the R or Python
-bindings to CmdStan, or can use the command `cmdstan_model`
-to activate the CmdStan makefile from anywhere. To enable either
-of these, ensure your environment is activated (via
-`conda activate cmdstan`, or whichever name you used above).
+The conda installation is designed so one can use the R or Python
+bindings to CmdStan seamlessly. Additionally, it provides the command
+`cmdstan_model` to activate the CmdStan makefile from anywhere.
+To enable either of these, ensure your environment is activated^[A special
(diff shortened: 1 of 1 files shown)


src/cmdstan-guide/installation.Rmd
```diff
@@ -1,6 +1,38 @@
...
+
+CmdStan can be installed via the package management system [conda](https://docs.conda.io/en/latest/)
+via the [conda-forge channel](https://conda-forge.org/). This
+will install a pre-built version of CmdStan along with the required
+dependencies (i.e. a C++ compiler, a version of Make, and required
+libraries) detailed below under [Source installation].
+
+One can create a new conda environment and install CmdStan with
+the following command.
+
```
bob-carpenter, line 13: I don't think there should be space here.

src/cmdstan-guide/installation.Rmd
```diff
@@ -1,6 +1,38 @@
...
+```
+conda create -n cmdstan -c conda-forge cmdstan
+```
+or install it in an existing environment with
+```
+conda install -c conda-forge cmdstan
+```
+
+With the conda installation, one can use the R or Python
+bindings to CmdStan, or can use the command `cmdstan_model`
```
bob-carpenter, line 23: ", or can use" -> "or"

It'd probably be better as two sentences.

src/cmdstan-guide/installation.Rmd
```diff
@@ -1,6 +1,38 @@
...
+via the [conda-forge channel](https://conda-forge.org/). This
+will install a pre-built version of CmdStan along with the required
+dependencies (i.e. a C++ compiler, a version of Make, and required
+libraries) detailed below under [Source installation].
+
+One can create a new conda environment and install CmdStan with
+the following command.
+
+```
+conda create -n cmdstan -c conda-forge cmdstan
```
bob-carpenter, line 15: I think this second "cmdstan" is a free name from the user that you call out later.  Either describe here that the second "cmdstan" can be anything and the subsequent command has to match.

src/cmdstan-guide/installation.Rmd
```diff
@@ -1,6 +1,38 @@
...
+conda install -c conda-forge cmdstan
+```
+
+With the conda installation, one can use the R or Python
+bindings to CmdStan, or can use the command `cmdstan_model`
+to activate the CmdStan makefile from anywhere. To enable either
+of these, ensure your environment is activated (via
+`conda activate cmdstan`, or whichever name you used above).
+
+Issues with the conda installation are handled separately.
```
bob-carpenter, line 28: I don't think our users will understand what separate means here.  So I'd suggest dropping line 28 and making line 29 more specific.  

```
Please report conda-specific install problems directly to the conda-forge issue tracker,
```

src/cmdstan-guide/installation.Rmd
```diff
@@ -1,6 +1,38 @@
...
+or install it in an existing environment with
+```
+conda install -c conda-forge cmdstan
+```
+
+With the conda installation, one can use the R or Python
+bindings to CmdStan, or can use the command `cmdstan_model`
+to activate the CmdStan makefile from anywhere. To enable either
+of these, ensure your environment is activated (via
+`conda activate cmdstan`, or whichever name you used above).
```
bob-carpenter, line 26: It helps to have a name for the name we're talking about here, like "package identifier" or whatever conda uses.  Then you can replace "name" with "package identifier" and it'll be clearer.  Then you can write `conda activate <package-id>` and say `<package-id>` is `cmdstan` or whatever other identifier you chose.

bob-carpenter (changes requested):
some very minor change suggestions

src/cmdstan-guide/installation.Rmd
WardBrian: it's actually the first cmdstan, which is even more reason to explain

src/cmdstan-guide/installation.Rmd
```diff
@@ -8,25 +8,30 @@ will install a pre-built version of CmdStan along with the required
...
 libraries) detailed below under [Source installation].
 
-One can create a new conda environment and install CmdStan with
-the following command.
-
+One can create a new conda environment (named `stan`)
+and install CmdStan with the following command.
 ```
-conda create -n cmdstan -c conda-forge cmdstan
+conda create -n stan -c conda-forge cmdstan
```
bob-carpenter, line 14: [optional]
Do you think `stan` might be confused here with our repo?  The repo is dev-facing and this is user-facing, so it should be OK.  I can't think of a better name than `stan` that isn't `cmdstan`.  Probably something like `cs` would be too confusible.

src/cmdstan-guide/installation.Rmd
```diff
@@ -8,25 +8,30 @@ will install a pre-built version of CmdStan along with the required
...
+bindings to CmdStan seamlessly. Additionally, it provides the command
+`cmdstan_model` to activate the CmdStan makefile from anywhere.
+To enable either of these, ensure your environment is activated^[A special
+environment, called `base`, is activated automatically by conda on startup.
+Installing software directly in `base` is not recommended by conda, but may be
+appropriate for some users who do not intend to use multiple environments]
+with `conda activate <env>`, where `<env>` is the name of
+the environment CmdStan was installed into. In the first command
+above, the environment was named `stan`, but any name may be used
+so long as it you are consistent.
```
bob-carpenter, line 31: as it you -> as you

Also, if `stan` is changed, the name here will have to change.

src/cmdstan-guide/installation.Rmd
bob-carpenter, line 14: [optional]
Would it be useful to remind the user how to check which version of cmdstan is installed?

bob-carpenter (changes requested):
A grammaro slipped through and then an optional question about whether `stan` might be confusible (probably not, which is why I marked it optional).

src/cmdstan-guide/installation.Rmd
WardBrian: I've seen `-env` as a suffix for this type of thing, so it would be `stan-env`. Longer than I would personally use, but pretty clear that it's not a magic word I think. Thoughts?

WardBrian: Something like this?

> 
> You can check which version of CmdStan is installed in your current environment by running 
> ```
> conda list cmdstan
> ```
> and you can specify a different version for installation by adding `=VERSION` to the end of the above commands.

bob-carpenter: That's great if you can do it through conda.  I think that'd help people verify they've actually got the version they want.

bob-carpenter: I like `-env` because it'll just reinforce that it's an environment in the docs.  I would probably just use `cmdstan` as you originally did (not in the docs, where it's a bit confusing, but in practice).

PR #403, opened 2021-09-17
minor_edit

CGMossa:


bob-carpenter (approved):
thanks!  no idea how the cdot crept into the dimension decls!

PR #418, opened 2021-10-12
Add "Available since version" to function reference

### LINKED ISSUE #417: add version provenance: (since [at least]) version x
#### Summary:

For the functions reference, add a provenance for each function indicating when it was introduced.   

#### Description:

This is especially important now that the interfaces are releasing asynchronously and users often have later doc than the version of Stan they're using.

#### Additional Information:

It will be OK to follow the Python docs and allow "since at least".

It would also be nice to add the same thing for language features, but it's less clear where to document them.  For example, the `complex` type was introduced recently, but types are described all over the doc.

There are lots of place this is listed: release notes, auto-generated function sigs, the actual stanc3 or previous stanc.

#### Current Version:
v2.28.0

WardBrian:
#### Submission Checklist

- [x] Builds locally 

#### Summary

This closes #417. Each function has had a version number added to it. 

This represents the minimum version one should expect the function to appear in _in its current form_, meaning renames or changes in return type/signature count as moving up a version. 

This was built from a list automatically generated from old documentation back to version 2.0, with some manual adjustments provided by @rok-cesnovar, @bob-carpenter, and by looking through old release notes. I'm fairly confident most of the version numbers are correct, and when they miss they miss _newer_, so we never say something was available during a version it wasn't, we just may be late here or there.

Here is how it looks in PDF:

Screenshot from 2021-10-12 14-49-31

And web:

Screenshot from 2021-10-12 14-49-58

#### Copyright and Licensing

Simons Foundation

### DIFF
--- a/src/functions-reference/complex-valued_basic_functions.Rmd
+++ b/src/functions-reference/complex-valued_basic_functions.Rmd
@@ -119,6 +119,7 @@ promoted to type `complex` before performing the operation.
 Return the complex argument `z`, \[ +z = z. \]
 `r since("2.28")`
 
+
 <!-- complex; operator-; (complex z); -->
 \index{{\tt \bfseries operator\_subtract}!{\tt (complex z): complex}|hyperpage}
 
@@ -212,6 +213,7 @@ Return 1 if x is equal to y and 0 otherwise,
 \]
 `r since("2.28")`
 
+
 <!-- int; operator!=; (complex x, complex y); -->
 \index{{\tt \bfseries operator\_logical\_not\_equal }!{\tt (complex x, complex y): int}|hyperpage}
 
@@ -243,6 +245,7 @@ imaginary component is 0.0.
 `void` **`operator=`**`(complex x, complex y)`<br>\newline
 `y = x;` assigns a (copy of) the value of `y` to `x`.
 `r since("2.28")`
+
 
 <!-- void; operator+=; (complex x, complex y); -->
 \index{{\tt \bfseries operator\_compound\_add }!{\tt (complex x, complex y): void}|hyperpage}
@@ -279,10 +282,10 @@ The following functions are specific to complex numbers other than
 absolute value, which has a specific meaning for complex numbers.
 
 
-<!-- complex; abs; (complex z); -->
-\index{{\tt \bfseries abs }!{\tt (complex z): complex}|hyperpage}
+<!-- real; abs; (complex z); -->
+\index{{\tt \bfseries abs }!{\tt (complex z): real}|hyperpage}
 
-`complex` **`abs`**`(complex z)`<br>\newline
+`real` **`abs`**`(complex z)`<br>\newline
 Return the absolute value of z, also known as the modulus or
 magnitude, which for $z = x + yi$ is
 \[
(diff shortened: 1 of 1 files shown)


src/functions-reference/real-valued_basic_functions.Rmd
```diff
@@ -294,65 +294,74 @@ Constants are represented as functions with no arguments and must be
 called as such.  For instance, the mathematical constant $\pi$ must be
 written in a Stan program as `pi()`.
 
-<!-- real; pi; () -->
+<!-- real; pi; (); -->
```
bob-carpenter, line 297: [comment]
more good catches.  thanks!

src/functions-reference/unbounded_continuous_distributions.Rmd
```diff
@@ -731,15 +806,16 @@ Generate a gumbel variate with location mu and scale beta; may only be
...
 
 ### Probability density function
 
 If $\mu \in \mathbb{R}$, $\sigma \in \mathbb{R}^+$ and $\tau \in [0, 1]$,
-then for $y \in \mathbb{R}$, 
-$$\begin{aligned} 
-& \text{SkewDoubleExponential}  (y|\mu,\sigma, \tau) = \\ 
+then for $y \in \mathbb{R}$,
+$$\begin{aligned}
+& \text{SkewDoubleExponential}  (y|\mu,\sigma, \tau) = \\
```
bob-carpenter, line 818: [optional]
I strongly prefer`\mid` over `|` as it puts a bit of space around the vertical bar.

Also, it's conventional to have the start/end math symbols (here `$$`) be on their own line.

It's also sub-optimal to have lines wrap---better to break up expressions into natural chunks.

The use of `I` should be `\textrm{I}` because that's how special functions get set in math.

I prefer to use `(`  everywhere instead of using `{` or `[` in some places.  If they get confusing to parse, either think about rewriting in smaller chunks or using `\left(` and `\right)`.

bob-carpenter (changes requested):
A question, some optional fixes depending on your patience, and a lot of gratitude for finding and fixing so many existing errors.  This doesn't necessarily need any fixes, but I'm still marking as requesting changes to wait for answers to the questions before merging.

src/functions-reference/complex-valued_basic_functions.Rmd
```diff
@@ -62,25 +62,25 @@ complex z3 = to_complex(re, im);  // z3 = -2.9 + 1.3i
 ```
 
 <!-- complex; to_complex; (); -->
-\index{{\tt\bfseries to\_complex }!{\tt (): complex}|hyperpage}
+\index{{\tt \bfseries to\_complex }!{\tt (): complex}|hyperpage}
```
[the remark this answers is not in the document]
WardBrian: There is a space in every other function definition, e.g.

https://github.com/stan-dev/docs/blob/c531196bca5e02fffec822b3e7128b454cef7519/src/functions-reference/integer-valued_basic_functions.Rmd#L86


Without the spaces here, things like `operator+` were showing up twice in the index. I think the latex index functionality is pretty specific when it comes to matches

src/functions-reference/unbounded_continuous_distributions.Rmd
WardBrian: This also only shows up in the diff due to whitespace removal. I agree with your suggestions, but they seem best in a separate PR

bob-carpenter:
I'm afraid merging the complex return types caused a conflict.  It should be good to go after that.

PR #448, opened 2021-12-02
automatically detect stan version numbers in function_sigs script

spinkney:
#### Summary

I got tired of putting in the Stan major and minor versions for the extract_function_sigs.py script. This looks at the index URL and extracts those from the redirecting link.

#### Copyright and Licensing

Sean Pinkey

### DIFF
--- a/extract_function_sigs.py
+++ b/extract_function_sigs.py
@@ -8,9 +8,7 @@ import os
 import os.path
 import sys
 import contextlib
-from lxml import html
-import requests
-import re
+import subprocess
 
 @contextlib.contextmanager
 def pushd(new_dir):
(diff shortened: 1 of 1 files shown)


extract_function_sigs.py
```diff
@@ -8,7 +8,9 @@
 import os.path
 import sys
 import contextlib
-
+from lxml import html
```
WardBrian, line 11: I don't think we need to make such a simple script have a dependency (outside of the python standard library)

extract_function_sigs.py
```diff
@@ -23,8 +25,16 @@ def main():
...
-        sys.exit(1)
+        page = requests.get('https://mc-stan.org/docs/functions-reference/index.html')
+        if page.status_code == 404:
+            print('Stan version not found! Add 2 arguments <MAJOR> <MINOR> version numbers')
+            sys.exit(1)
+            
+        tree = html.fromstring(page.content)  
+        stan_url = tree.xpath('/html/body/a/@href')
+        x = re.findall(r'[0-9]+', stan_url[0])
+        stan_major = int(x[0])
```
WardBrian, line 36: I think this whole else block should be wrapped in a try/except - it's definitely conceivable that the above page doesn't 404 but the regex fails, for example.

WardBrian (commented):
A couple notes.

I also don't think the version is used for anything other than the file name, correct? If so I think it's reasonable (maybe even better) to tag it with the current git hash or just `develop`.

extract_function_sigs.py
spinkney: > A couple notes.
> 
> I also don't think the version is used for anything other than the file name, correct? If so I think it's reasonable (maybe even better) to tag it with the current git hash or just `develop`. For example, if you run the script against master, functions may have been added since the last push online.

It's just the name. Yea, the current git hash would be good.

spinkney:
@WardBrian uses the current branch HEAD as the hash for the file. If the git command does not work raises an error to input versions.

WardBrian (approved):

WardBrian:
Thanks!

PR #461, opened 2022-01-07
Meta analysis log-odds code is wrong

spinkney:
There needs to be a parenthesis around the subtraction in the transformed data block.

### DIFF
--- a/src/stan-users-guide/measurement-error.Rmd
+++ b/src/stan-users-guide/measurement-error.Rmd
@@ -285,7 +285,7 @@ transformed data {
   array[J] real<lower=0> sigma;
   for (j in 1:J) {
     y[j] = log(r_t[j]) - log(n_t[j] - r_t[j])
-            - ((log(r_c[j]) - log(n_c[j] - r_c[j]));
+            - (log(r_c[j]) - log(n_c[j] - r_c[j]));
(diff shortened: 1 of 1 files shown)


src/stan-users-guide/measurement-error.Rmd
```diff
@@ -285,7 +285,7 @@ transformed data {
   array[J] real<lower=0> sigma;
   for (j in 1:J) {
     y[j] = log(r_t[j]) - log(n_t[j] - r_t[j])
-            - (log(r_c[j]) - log(n_c[j] - r_c[j]);
+            - ((log(r_c[j]) - log(n_c[j] - r_c[j]));
```
andrjohns, line 288: Shouldn't this only have a single parenthesis added, otherwise it's still unbalanced?

```suggestion
            - (log(r_c[j]) - log(n_c[j] - r_c[j]));
```

andrjohns (changes requested):

src/stan-users-guide/measurement-error.Rmd
spinkney: ah yes 👍

andrjohns (approved):
Great, thanks!

PR #473, opened 2022-01-14
Document function overloading

WardBrian:
#### Submission Checklist

- [x] Builds locally 

#### Summary

Documentation for https://github.com/stan-dev/stanc3/pull/1027

#### Copyright and Licensing

Simons Foundation

### DIFF
--- a/src/reference-manual/user-functions.Rmd
+++ b/src/reference-manual/user-functions.Rmd
@@ -39,9 +39,8 @@ types `real` and `real`.  The arguments are named `mu`
 and `sigma`, but that is not part of the declaration.
 
 ### Function overloading
-Two user-defined functions may have the same name if they have
+Multiple user-defined functions may have the same name if they have
 different sequences of [argument types](#argument-types-and-qualifiers).
-
 This is known as function overloading.
 
 For example, the following two functions are both defined with the name `add_up`
@@ -56,18 +55,23 @@ real add_up(real a, real b, real c){
 }
 ```
 
-The return types of overloaded functions do not need to be the same as a matter
-of course, but it may be confusing if they differ.
+The return types of overloaded functions do not need to be the same. One could define an additional `add_up` function as follows
 
-Two function may **not** use the same name if their signature _only_ differs by the
+```stan
+int add_up(int a, int b){
+  return a + b;
+}
+```
+
+That being said, functions may **not** use the same name if their signature _only_ differs by the
 return type.
 
-For example,
+For example, the following is not permitted
 ```stan
+// illegal
 real baz(int x);
 int baz(int x);
 ```
-is not permitted.
 
 Function names used in the Stan standard library may be overloaded by
 user-defined functions. Exceptions to this are the `reduce_sum` family of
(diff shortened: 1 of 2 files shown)


src/reference-manual/user-functions.Rmd
```diff
@@ -36,10 +36,42 @@ real foo(real mu, real sigma);
 
 declares a function named `foo` with two argument variables of
 types `real` and `real`.  The arguments are named `mu`
-and `sigma`, but that is not part of the declaration.  Two
-user-defined functions may *not* have the same name even if they have
-different sequences of argument types.
+and `sigma`, but that is not part of the declaration.
 
+### Function overloading
+Two user-defined functions may have the same name if they have
```
bob-carpenter, line 42: Two -> More than one

src/reference-manual/user-functions.Rmd
```diff
@@ -36,10 +36,42 @@ real foo(real mu, real sigma);
...
+  return a + b;
+}
+
+real add_up(real a, real b, real c){
+  return a + b + c;
+}
+```
+
+The return types of overloaded functions do not need to be the same as a matter
+of course, but it may be confusing if they differ.
```
bob-carpenter, line 60: We don't have dependent types, so we'll run into cases most naturally coded with differing returns, as in 

```
int add_up(int a, int b) { return a + b; }
```

src/stan-users-guide/user-functions.Rmd
```diff
@@ -404,10 +404,72 @@ function (`_lpmf`) must be discrete (integer or integer array).
...
+
+Furthermore, there must be only one such signature, e.g., the minimum number of
+promotions must be a unique minimum. This requirement does forbid certain kinds
+of overloading. For example, the function signatures
+```stan
+real bar(int x, real y);
+real bar(real x, int y);
+```
+These signatures do not have a unique minimum number of promotions for the call
+`bar(1, 2)`. Both signatures require one promotion, and so it cannot be
```
bob-carpenter, line 457: [comment]
Nice---this is the form I was suggesting before.

bob-carpenter (changes requested):
This is a ton of minor comments, largely about when to provide paragraphs and how to separate displays.  Hope they're not too painful.

src/reference-manual/user-functions.Rmd
```diff
@@ -36,10 +36,42 @@ real foo(real mu, real sigma);
...
+For example,
+```stan
+real baz(int x);
+int baz(int x);
+```
+is not permitted.
+
+Function names used in the Stan standard library may be overloaded by
+user-defined functions. Exceptions to this are the `reduce_sum` family of
+functions and ODE integrators, which cannot be overloaded.
```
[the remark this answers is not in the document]
WardBrian: This is not strictly true - users can override `map_rect`, for example. 

The true restriction is actually that UDFs cannot override variadic functions, as these are currently typechecked entirely separately based on their name. I can say this if we want, but I was trying to avoid the word "variadic"

bob-carpenter (approved):
That was quick. Thanks!

PR #531, opened 2022-05-17
Document abs/fabs changes

### LINKED ISSUE #523: doc for `R` and `T` arg/return types
#### Summary:

In the Functions Reference, document the argument and return type restrictions in every polymorphically typed function.

#### Descriptions

For example, the doc for `fabs` looks like this.

```
R fabs(T x)
absolute value of x
Available since 2.0, vectorized in 2.13
```

with no documentation around what `R` and `T` are.  

Also, the `abs` doc is wrong because it should have return type `T` for argument type `T` (i.e., not the same as `fabs`).

Add documentation for all of these argument types and return types.  

#### Additional Information:

One way to do it would be to add a generic description somewhere and then link to it in each function description.

#### Current Version:
v2.18.0

WardBrian:
#### Submission Checklist

- [x] Builds locally 

#### Summary

Documentation for https://github.com/stan-dev/stanc3/pull/1195. Closes #523.

`abs` is now vectorized. The previous note about it being deprecated for real arguments is removed, `fabs` is now deprecated instead.

#### Copyright and Licensing

Simons Foundation

### DIFF
--- a/src/functions-reference/integer-valued_basic_functions.Rmd
+++ b/src/functions-reference/integer-valued_basic_functions.Rmd
@@ -135,12 +135,13 @@ This is a no-op. \[ \text{operator+}(x) = x \]
 
 ## Absolute functions
 
-<!-- R; abs; (T x); -->
-\index{{\tt \bfseries abs }!{\tt (T x): R}|hyperpage}
+<!-- T; abs; (T x); -->
+\index{{\tt \bfseries abs }!{\tt (T x): T}|hyperpage}
 
-`R` **`abs`**`(T x)`<br>\newline
-absolute value of x
-`r since("2.0")`
+`T` **`abs`**`(T x)`<br>\newline
+The absolute value of x. This function works elementwise over containers such as
+arrays.
+`r since("2.0, vectorized in 2.30")`
 
 <!-- int; int_step; (int x); -->
 \index{{\tt \bfseries int\_step }!{\tt (int x): int}|hyperpage}
@@ -192,4 +193,3 @@ Return the maximum of x and y. \[ \text{max}(x, y) = \begin{cases} x &
 
 Return the size of `x` which for scalar-valued `x` is 1
 `r since("2.26")`
-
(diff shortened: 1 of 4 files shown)


src/functions-reference/integer-valued_basic_functions.Rmd
```diff
@@ -135,12 +135,13 @@ This is a no-op. \[ \text{operator+}(x) = x \]
...
-<!-- R; abs; (T x); -->
-\index{{\tt \bfseries abs }!{\tt (T x): R}|hyperpage}
+<!-- T; abs; (T x); -->
+\index{{\tt \bfseries abs }!{\tt (T x): T}|hyperpage}
 
-`R` **`abs`**`(T x)`<br>\newline
-absolute value of x
-`r since("2.0")`
+`T` **`abs`**`(T x)`<br>\newline
+The absolute value of x. This function works elementwise over containers such as
```
bob-carpenter, line 142: I'd like to see `T` described as applying to any real or integer value or real or integer-valued container.    As is, it never says it can apply to `real` or `int`.

src/functions-reference/real-valued_basic_functions.Rmd
```diff
@@ -687,12 +687,20 @@ used in the   data, transformed data, or generated quantities blocks.*
 
 ### Absolute value functions
 
+<!-- T; abs; (T x); -->
+\index{{\tt \bfseries abs }!{\tt (T x): T}|hyperpage}
+
+`T` **`abs`**`(T x)`<br>\newline
+The absolute value of x. This function works elementwise over containers such as
```
bob-carpenter, line 694: This is also confusing because this signature only applies to real or integer values or real- or integer-valued containers.  When the argument is `complex`, it's a different type.  So we need to say that.

bob-carpenter (changes requested):
small change around defining type constraints more tightly and being sensitive to `complex` existing

bob-carpenter (approved):
thanks for fixing---I think it'll be much less confusing this way

PR #557, opened 2022-07-05
revised thinning section

bob-carpenter:
#### Submission Checklist

- [x] Builds locally 

#### Summary

Revise section on thinning to highlight anti-correlated draws as a possibility.

#### Copyright and Licensing

Please list the copyright holder for the work you are submitting (this will be you or your assignee, such as a university or company):  Simons Foundation

### DIFF
--- a/src/reference-manual/analysis.Rmd
+++ b/src/reference-manual/analysis.Rmd
@@ -352,11 +352,16 @@ N_{\mathrm{eff}}
 \frac{N}{1 + 2 \sum_{t = 1}^{\infty} \rho_t}.
 $$
 
-Effective sample size $N_{\mathrm{eff}}$ can be larger than $N$ in
-case of antithetic Markov chains, which have negative autocorrelations
-on odd lags. The no-U-turn sampling (NUTS) algorithm used in Stan can
-produce $N_{\mathrm{eff}}>N$ for parameters which have close to
-Gaussian posterior and little dependency on other parameters.
+
+For independent draws, the effective sample size is just the number of
+iterations.  For correlated draws, the effective sample size will be
+lower than the number of iterations.  For anticorrelated draws, the
+effective sample size can be larger than the number of iterations.  In
+this latter case, MCMC can work better than independent sampling for
+some estimation problems.  Hamiltonian Monte Carlo, including the
+no-U-turn sampler used by default in Stan, can produce anticorrelated
+draws if the posterior is close to Gaussian with little posterior
+correlation.
 
 
 ### Estimation of effective sample size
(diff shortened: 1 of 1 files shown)


betanalpha:
I know that this has already been merged but I wanted to comment on some misunderstandings in the language.

Antithethic Markov chains — where one parameter oscillates between two distinct regions after each transition — are the result of _periodic_ Markov transitions.  In a periodic Markov transition Markov chain Monte Carlo estimators may converge to target expectation values but the N-step distribution will not converge to the target distribution.  This means that applying the Markov transition once may not have the invariant distribution as applying the Markov transition twice or more.  Because of this thinning can actually change the invariant distribution, and hence all of the asymptotic behavior of Markov chain Monte Carlo estimators.

Every Hamiltonian Monte Carlo transition, however, is aperiodic because of the always non-zero probability of staying at the initial point.  Markov chains might oscillate between two regions with high probability but not deterministically, which is fundamentally different from antithetic behavior.  Consequently we should really avoid referring to “antithetical” anywhere in the documentation.

Correlations of Markov chain states and effective sample size of Markov chain Monte Carlo estimators cannot be spoken of monolithically — they are defined for each expectand of interest.  In general each parameter function will have different autocorrelations and hence different effective sample sizes.  Same for any nontrivial function of the parameters.  What we can say is that if the incremental effective sample for a given expectand f is less than one,

ESS_{f} / I < 1

then thinning should result in a slightly larger incremental effective sample size but a smaller total effective sample size _for that expectand_.  If ESS_{f} / I > 1 then thinning should decrease both, again for that particular expectand.  In general thinning the same Markov chain can have different effects on the incremental and total effective sample sizes for different expectands.

bob-carpenter:
Thanks, @betanalpha.  

I was using "antithetical" following its informal use in MCMC (as in the top 10 papers that are hits for that search term on Google).  I didn't realize there was a more precise mathematical meaning.  I'm totally OK fixing this to be more precise, but I think it needs to be connected back to the informal usage of "antithetical" in the field even if only to say people are using the word wrong.  That's what Andrew did for "burn in" and we changed to saying "warmup" instead.

Do you want to create a PR to fix?

betanalpha:
> I was using "antithetical" following its informal use in MCMC (as in the top 10 papers that are hits for that search term on Google). I didn't realize there was a more precise mathematical meaning. I'm totally OK fixing this to be more precise, but I think it needs to be connected back to the informal usage of "antithetical" in the field even if only to say people are using the word wrong.
> 
To be clear the typical use of antithetical to mean “Markov transition oscillates between positive and negative values” is not a problem.  The complication is that by construction Stan’s Hamiltonian Monte Carlo sampler will never exhibit exact antithetical behavior (because the probability of staying at the initial point is always non-zero).

The practical consequence is that thinning can increase the Monte Carlo standard error, as correctly implied in the current text, but it can’t obstruct convergence as thinning can do to a Markov transition that is actually antithetical (or more formally periodic with a periodicity of 2).  In other words there’s a meaningful difference between a Markov transition that is actually antithetical and one that just exhibits anti-correlations.

This subtlety can be avoided entirely by saying something like “When Markov chains generated from Stan’s Hamiltonian Monte Carlo sampler exhibit strong anti-correlations then thinning will remove information and actually increase the error of expectation value estimators.  In this case thinning should be avoided entirely.”, avoiding reference to “antithetical” entirely.  I think that trying to connect this advice to the greater folklore around antithetical Markov chains will open to big of a whole for this small piece of doc.

PR #592, opened 2022-11-03
documentation for algebra solver with variadic argument.

charlesm93:
#### Submission Checklist

- [ ] Builds locally* 

*: builds mostly. I added two references which don't seem to render (i.e. I pick up question marks)

#### Summary

Documents the algebraic solvers with variadic arguments. Moves the original algebraic solvers to the section on deprecated functions. 

#### Copyright and Licensing

Please list the copyright holder for the work you are submitting (this will be you or your assignee, such as a university or company): Flatiron Institute

### DIFF
--- a/src/functions-reference/higher-order_functions.Rmd
+++ b/src/functions-reference/higher-order_functions.Rmd
@@ -44,13 +44,29 @@ The signature of the algebraic system is then:
  vector algebra_system (vector y, ...)
 ```
 
+There is no type restriction for the variadic arguments
+and each argument can be passed as data or parameter.
+However users should use parameter arguments only when nescessary
+and mark data arguments with the keyword `data`.
+In the below example, the last variadic argument, $x$, is restricted to being data:
+
+```stan
+  vector algebra_system (vector y, vector theta, data vector x)
+```
+
+Distinguishing data and parameter is important for computational reasons.
+Augmenting the total number of parameters increases the cost of propagating derivatives
+through the solution to the algebraic equation,
+and ultimately the computational cost of evaluating the gradients.
+
+
 ### Call to the algebraic solver
 
 `vector` **`solve_newton`**`(function algebra_system, vector y_guess, ...)`<br>\newline
 Solves the algebraic system, given an initial guess, using Newton's method.
 `r since("2.31")`
 
-`vector` **`solve_newton_tol`**`(function algebra_system, vector y_guess, data real scaling_step, data real f_tol, int max_steps)`<br>\newline
+`vector` **`solve_newton_tol`**`(function algebra_system, vector y_guess, data real scaling_step, data real f_tol, int max_steps, ...)`<br>\newline
 Solves the algebraic system, given an initial guess, using Newton's method with additional control parameters for the solver.
 `r since("2.31")`
 
@@ -58,7 +74,7 @@ Solves the algebraic system, given an initial guess, using Newton's method with 
 Solves the algebraic system, given an initial guess, using Powell's hybrid method.
 `r since("2.31")`
 
-`vector` **`solve_powell_tol`**`(function algebra_system, vector y_guess, data real rel_tol, data real f_tol, int max_steps)`<br>\newline
+`vector` **`solve_powell_tol`**`(function algebra_system, vector y_guess, data real rel_tol, data real f_tol, int max_steps, ...)`<br>\newline
 Solves the algebraic system, given an initial guess, using Powell's hybrid method with additional control parameters for the solver.
 `r since("2.31")`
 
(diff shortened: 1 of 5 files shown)


src/functions-reference/higher-order_functions.Rmd
```diff
@@ -7,173 +7,148 @@ by name to the higher-order functions.
...
 another function as one of its arguments. Other functions in
-Stan which share this feature are the ordinary differential equation
-solvers (see section [Ordinary Differential Equation (ODE) Solvers](#functions-ode-solver)).
+Stan which share this feature are the differential equation
+solvers (see section [Ordinary Differential Equation (ODE) Solvers](#functions-ode-solver) and
+[Differential Algebraic Equation (DAE) solver](#functions-dae-solver)).
 Ordinary Stan functions do not allow functions as arguments.
 
-Second, some of the arguments of the algebraic solvers are restricted
-to data only expressions. These expressions must not contain variables
```
WardBrian, line 34: While it is correct to delete this paragraph, it might be good to mention the computational advantages of marking arguments as `data` when needed in the discussion of the variadic arguments.

src/functions-reference/higher-order_functions.Rmd
```diff
@@ -7,173 +7,148 @@ by name to the higher-order functions.
...
-\index{{\tt \bfseries algebra\_solver }!{\tt (function algebra\_system, vector y\_guess, vector theta, data  array[] real x\_r, array[] int x\_i, data real rel\_tol, data real f\_tol, int max\_steps): vector}|hyperpage}
+`vector` **`solve_newton`**`(function algebra_system, vector y_guess, ...)`<br>\newline
+Solves the algebraic system, given an initial guess, using Newton's method.
+`r since("2.31")`
 
-`vector` **`algebra_solver`**`(function algebra_system, vector y_guess, vector theta, data array[] real x_r, array[] int x_i, data real rel_tol, data real f_tol, int max_steps)`<br>\newline
-Solves the algebraic system, given an initial guess, using the Powell
-hybrid algorithm with additional control parameters for the solver.
-`r since("2.17")`
+`vector` **`solve_newton_tol`**`(function algebra_system, vector y_guess, data real scaling_step, data real f_tol, int max_steps)`<br>\newline
```
WardBrian, line 53: Both `_tol` signatures are missing the `...`

src/functions-reference/higher-order_functions.Rmd
charlesm93: Ah, I didn't realize arguments had to be explicitly marked using `data` to avoid propagating derivatives with respect to them (I guess that makes sense). Here's an additional paragraph:


There is no type restriction for the variadic arguments
and each argument can be passed as data or parameter.
However users should use parameter arguments only when nescessary,
and mark data arguments with the keyword `data`.
In the below example, the last variadic argument, $x$, is restricted to being data:

```stan
  vector algebra_system (vector y, vector theta, data vector x)
```

Distinguishing data and parameter is important for computational reasons.
Augmenting the total number of parameters increases the cost of propagating derivatives
through the solution to the algebraic equation,
and ultimately the computational cost of evaluating the gradient of the target.

src/functions-reference/higher-order_functions.Rmd
charlesm93: Good catch. Fixed!

charlesm93:
Also any idea why the references I added at the end of the section don't render?

WardBrian:
The reference renders for me. Sometimes the references require you to build multiple times in a row (first build generates the references, second build picks them up).

WardBrian:
Can you also update `src/reference-manual/expressions.Rmd` and (more optionally) `src/stan-users-guide/algebraic-equations.Rmd`?

The only other mention of the old `algebra_solver` is a somewhat confusing section in `src/reference-manual/syntax.Rmd` which I think should probably just be deleted.

charlesm93:
Regarding `syntax.Rmd` it looks like there is also a section on the deprecated integrated_ode functions. I suppose the section on algebra_solver can be deleted, but if so, would it also make sense to delete the section on integrate_ode?

WardBrian:
I personally don't think that the subheadings for the algebra solver, 1d integrator, or ode solvers really belong on that page. I think in stanc2 those were implemented as core parts of the language so it sort of made sense, but in stanc3 they're just functions and shouldn't be treated as special cases in the reference manual IMO.

bob-carpenter:
I agree that rearranging the manual to fold these in as regular functions makes sense.

charlesm93:
I agree with you. Let me just check in with folks very quickly on the stan forum and if tomorrow we don't hear any objections, we'll move forward.

charlesm93:
Ok. I'll remove the section on higher-order functions from language syntax. It's redundant with what's written in other chapters of the manual.

WardBrian (approved):
LGTM! 
Thanks for taking point on this @charlesm93

PR #664, opened 2023-08-16
Add documentation on tuple-returning special functions

WardBrian:
#### Submission Checklist

- [x] Builds locally 

#### Summary

Docs for https://github.com/stan-dev/math/pull/2931 and https://github.com/stan-dev/stanc3/pull/1346

#### Copyright and Licensing

Simons Foundation

### DIFF
--- a/src/functions-reference/complex_matrix_operations.Rmd
+++ b/src/functions-reference/complex_matrix_operations.Rmd
@@ -1342,6 +1342,15 @@ The matrix with the complex-valued (column) eigenvectors of the matrix `A` in th
 same order as returned by the function `eigenvalues`
 `r since("2.32")`
 
+<!-- tuple(complex_matrix, complex_vector); eigendecompose; (complex_matrix A); -->
+\index{{\tt \bfseries eigendecompose }!{\tt (complex\_matrix A): tuple(complex\_matrix, complex\_vector)}|hyperpage}
+
+`tuple(complex_matrix, complex_vector)` **`eigendecompose`**`(complex_matrix A)`<br>\newline
+Return the matrix of (column) eigenvectors and vector of eigenvalues of the
+matrix `A`. This function is equivalent to `(eigenvectors(A), eigenvalues(A))`
+but with a lower computational cost due to the shared work between the two
+results.
+`r since("2.33")`
 
 <!-- complex_vector; eigenvalues_sym; (complex_matrix A); -->
 \index{{\tt \bfseries eigenvalues\_sym }!{\tt (complex\_matrix A): complex\_vector}|hyperpage}
@@ -1357,6 +1366,16 @@ The vector of eigenvalues of a symmetric matrix `A` in ascending order
 The matrix with the (column) eigenvectors of symmetric matrix `A` in the
 same order as returned by the function `eigenvalues_sym`
 `r since("2.30")`
+
+<!-- tuple(complex_matrix, complex_vector); eigendecompose_sym; (complex_matrix A); -->
+\index{{\tt \bfseries eigendecompose\_sym }!{\tt (complex\_matrix A): tuple(complex\_matrix, complex\_vector)}|hyperpage}
+
+`tuple(complex_matrix, complex_vector)` **`eigendecompose_sym`**`(complex_matrix A)`<br>\newline
+Return the matrix of (column) eigenvectors and vector of eigenvalues of the
+symmetric matrix `A`. This function is equivalent to `(eigenvectors_sym(A),
+eigenvalues_sym(A))` but with a lower computational cost due to the shared work
+between the two results.
+`r since("2.33")`
 
 Because multiplying an eigenvector by $-1$ results in an eigenvector,
 eigenvectors returned by a decomposition are only identified up to a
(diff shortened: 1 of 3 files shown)


src/functions-reference/complex_matrix_operations.Rmd
```diff
@@ -1441,6 +1472,28 @@ Compute the unitary matrix of the complex Schur decomposition of `A`.
 `r since("2.31")`
 
 
+<!-- tuple(complex_matrix, complex_matrix); complex_schur_decompose; (matrix A); -->
+\index{{\tt \bfseries complex\_schur\_decompose }!{\tt (matrix A): tuple(complex\_matrix, complex\_matrix)}|hyperpage}
```
bob-carpenter, line 1476: Is there a reason this isn't just a promotion of the complex case?  Is there a different algorithm from Eigen?

src/functions-reference/matrix_operations.Rmd
```diff
@@ -2230,6 +2230,17 @@ The matrix with the complex-valued (column) eigenvectors of the matrix `A` in th
 same order as returned by the function `eigenvalues`
 `r since("2.30")`
 
+
+<!-- tuple(complex_matrix, complex_vector); eigendecompose; (matrix A); -->
```
bob-carpenter, line 2234: I think this needs to go on the complex_matrix_operations section if any complex component pollutes like for Schur return for real inputs.

src/functions-reference/matrix_operations.Rmd
```diff
@@ -2278,6 +2300,17 @@ implies that the resulting matrix is square with the same number of
 columns as A
 `r since("2.18")`
 
+
+<!-- tuple(matrix, matrix); qr_thin; (matrix A); -->
+\index{{\tt \bfseries qr\_thin }!{\tt (matrix A): tuple(matrix, matrix)}|hyperpage}
+
+`tuple(matrix, matrix)` **`qr_thin`**`(matrix A)`<br>\newline
+Returns both portions of the QR decomposition of A. This function
```
bob-carpenter, line 2308: This should mention that Q is orthonormal (in the column vectors) and R is upper triangular.

src/functions-reference/matrix_operations.Rmd
```diff
@@ -2359,6 +2392,18 @@ The left-singular vectors of `A`
 The right-singular vectors of `A`
 `r since("2.26")`
 
+<!-- tuple(matrix, vector, matrix); svd; (matrix A); -->
+\index{{\tt \bfseries svd }!{\tt (matrix A): tuple(matrix, vector, matrix)}|hyperpage}
+
+`tuple(matrix, vector, matrix)` **`svd`**`(matrix A)`<br>\newline
```
bob-carpenter, line 2398: I'm on the fence as to whether all these variants should say what the function does or just point out to other defs.

src/functions-reference/sparse_matrix_operations.Rmd
```diff
@@ -86,6 +86,16 @@ followed by the size of `csr_extract_w(a)` plus one; see section
 [compressed row storage](#CSR).
 `r since("2.8")`
 
+
+<!-- tuple(vector, array[] int, array[] int); csr_extract; (matrix a); -->
+\index{{\tt \bfseries csr\_extract }!{\tt (matrix a): tuple(vector, array[] int, array[] int)}|hyperpage}
+
+`tuple(vector, array[] int, array[] int)` **`csr_extract`**`(matrix a)`<br>\newline
+Return all three components of the CSR representation of the matrix `a`; see
```
bob-carpenter, line 94: Same comment---this doesn't really say what the function does directly.

bob-carpenter (changes requested):
This looks good and I can be convinced that we don't want to duplicate what the function does, but the shape of returns should be indicated, I think.  At least it should be consistent.

src/functions-reference/complex_matrix_operations.Rmd
WardBrian: The C++ implementation in `stan::math` is happy with either complex matrices or real ones. Eigen internally does template on whether or not the scalars are complex to optimize which routine is used.

It's important we expose the extra signature in the compiler to avoid the unnecessary promotion, but we don't need to document it if you'd prefer we didn't

src/functions-reference/matrix_operations.Rmd
WardBrian: I don't know if I'm super happy with the way we have these pages divided up, the reasoning behind putting the complex-return eigendecomposition functions here is that the `_sym` functions already were, and if someone is looking at those but wants a more general function it seems odd to have on a different page.

src/functions-reference/sparse_matrix_operations.Rmd
WardBrian: These functions all mainly link to the main page on CSR - `extract_w` just says "Return non-zero values in matrix a; see section `[compressed row storage](#CSR)`."

src/functions-reference/complex_matrix_operations.Rmd
bob-carpenter: In that case, I'd suggest adding to the doc a note that it's more efficient for real-valued inputs.

src/functions-reference/matrix_operations.Rmd
bob-carpenter: OK.  I agree that it's a bit of a mess.  I was the one who created the original taxonomy here and am totally OK with changing it if you have a better suggestion.  That doesn't need to happen in this PR, of course.

src/functions-reference/sparse_matrix_operations.Rmd
bob-carpenter: If you think that's OK, I'm OK leaving these things indirect.  I'm personally a bit frustrated with doc that sends me reference chasing, especially if they are on different pages of html.

bob-carpenter (approved):
thanks!

PR #668, opened 2023-09-05
Docs generated by Jenkins for v2-33

stan-buildbot:
Docs generated through the [Jenkins Job](https://jenkins.flatironinstitute.org/job/Stan/job/BuildDocs/)


WardBrian (approved):
This looks good. The generated HTML files indicate that we're using a newer version of bookdown (0.33 vs 0.22), but it doesn't look like any file names changed this time so I don't think that is an issue like it was going from 0.18 to 0.22

PR #687, opened 2023-11-06
Issue/686 - Rmd rendering in bookdown and quarto 

mitzimorris:
#### Submission Checklist

- [x] Builds locally 

#### Summary

changed use of `\[` and `/]` to `\begin{equation*}` and `\end{equation*}` so that Rmd files can render using either bookdown or quarto.

#### Copyright and Licensing


mitzimorris:
This is a work in progress - but as it will touch a lot of files, I wanted y'all to get a sense of what I propose to do here.

At this point, I have changed the Rmd files for the functions reference and verified that the HTML can render using either quarto or bookdown (i.e., current build scripts).    I also found a few omissions and fixed a couple of rendering issues.

mitzimorris:
fixed problems with build - now builds both pdf and html on my machine.
I've skimmed both versions, checking that equations look OK and that fit on the fixed pdf page width.
@WardBrian - let me know if it builds for you.

next up, figuring out how to build functions ref pdf using quarto - unless we want to recombine all ref manuals into a single pdf.

WardBrian:
Builds now, thanks!

mitzimorris:
This is ready for review - all Rmd files now render correctly as quarto html pages and as bookdown html and pdfs.

I also used quarto to build a pdf book version of the functions reference, which turned up a few stray miscodings of the index entries.   The other documents don't include an index.  To build a pdf with quarto, I needed to create a `_quarto.yml`.   I will check that work into a new sub-branch - I think that it would be best to merge this set of cleanups first.

@WardBrian - do you have cycles for this?

WardBrian:
Maybe not tomorrow but should be able to get to it before Thanksgiving certainly

WardBrian (approved):
Looks good - I needed to re-install bookdown to get the build working, so maybe the README should mention a newer version than 0.23. Since the eventual goal is to move away from bookdown entirely I don't think we need to be particularly conservative with requiring older versions or anything

PR #699, opened 2024-01-12
Update integrate_1d doc

avehtari:
- Mention that integrate_1d uses double exponential quadrature
- Mention what happens if relative_tolerance is not reached


bob-carpenter (approved):
Thanks.  Looks great.

PR #678, opened 2023-10-06
Document binomial-logit GLM

andrjohns:
#### Submission Checklist

- [x] Builds locally 
- [x] New functions marked with `` `r since("VERSION")` ``
- [x] Declare copyright holder and open-source license: see below

#### Summary

This PR adds documentation for the `binomial_logit_glm` GLM distribution added in [this PR](https://github.com/stan-dev/math/pull/2946). The implementation & likelihood/gradients are (unsurprisingly) very similar to the `bernoulli_logit_glm` distribution, so I've based the documentation on that entry.

Let me know if I've missed anything!

#### Copyright and Licensing

Please list the copyright holder for the work you are submitting (this will be you or your assignee, such as a university or company): Andrew Johnson

By submitting this pull request, the copyright holder is agreeing to license the submitted work under the following licenses:
- Code: BSD 3-clause (https://opensource.org/licenses/BSD-3-Clause)
- Documentation: CC BY-ND 4.0 (https://creativecommons.org/licenses/by-nd/4.0/)

PR #702, opened 2024-01-16
Add section documenting tuple unpacking

WardBrian:
#### Submission Checklist

- [x] Builds locally 

#### Summary

Docs for https://github.com/stan-dev/stanc3/pull/1360. Unfortunately I dropped the ball on documenting this earlier, so we will need to re-build the docs for the current release after this (and hopefully #678)

#### Copyright and Licensing

Simons Foundation

### DIFF
--- a/src/reference-manual/types.Rmd
+++ b/src/reference-manual/types.Rmd
@@ -1388,39 +1388,41 @@ corresponding container type promotions).
 
 ### Unpacking assignment of tuples {-}
 
-For convience of using values stored in tuples, Stan supports
+For convenience of using values stored in tuples, Stan supports
 "unpacking" (or "destructuring") of tuples in an assignment statement.
 
-Given a tuple `T` of type `tuple(T1, ..., Tn)` and a sequence of
-variables of types `v1`, ..., `vn`, where each `vi` has a type which is
-[assignable](#assignment-typing) from type `Ti`, individual elements of
-the tuple may be assigned to the corresponding variables in the sequence
-by the statement
+Given a tuple `t` of type `tuple(T1, ..., Tn)` and a sequence of
+[assignable expressions](#right-hand-side-expressions) of types
+`v1`, ..., `vn`, where each `vi` has a [type which is assignable](#assignment-typing)
+from type `Ti`, individual elements of the tuple may be assigned to the
+corresponding variables in the sequence by the statement
 
 ```stan
-(v1, /*...*/, vn) = T;
+(v1, /*...*/, vn) = t;
 ```
 
-For example, if `T` is a tuple of type `tuple(int, real, complex)`,
-then the program
-
-```stan
-int i;
-real r;
-complex z;
-
-(i, r, z) = T;
-```
-
-Assigns the result of `T.1` to `i`, the result of `T.2` to `r`, and
-the result of `T.3` to `z`.
+Note that the above parenthesis are required, unlike in some other
+languages with similar features (e.g., Python).
 
 These unpacking assignments can be nested if the tuple on the right hand
 side contains nested tuples.
 
-The left hand side of an unpacking assignment must contain only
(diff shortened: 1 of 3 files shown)


rok-cesnovar (approved):

src/reference-manual/types.Rmd
```diff
@@ -1386,6 +1386,43 @@ As with other assignments, promotions will happen if necessary (of
 `int` to `real` and of `real` to `complex`, along with the
 corresponding container type promotions).
 
+### Unpacking assignment of tuples {-}
+
+For convience of using values stored in tuples, Stan supports
```
bob-carpenter, line 1391: spelling: convenience

src/reference-manual/types.Rmd
```diff
@@ -1386,6 +1386,43 @@ As with other assignments, promotions will happen if necessary (of
...
+"unpacking" (or "destructuring") of tuples in an assignment statement.
+
+Given a tuple `T` of type `tuple(T1, ..., Tn)` and a sequence of
+variables of types `v1`, ..., `vn`, where each `vi` has a type which is
+[assignable](#assignment-typing) from type `Ti`, individual elements of
+the tuple may be assigned to the corresponding variables in the sequence
+by the statement
```
bob-carpenter, line 1398: Would it help to say these aren't general lvalues but must be indexes?  That is, we can't have indexed expressions on the LHS can we?

Would it also help to say that the parens are required, unlike in Python?

src/reference-manual/types.Rmd
```diff
@@ -1386,6 +1386,43 @@ As with other assignments, promotions will happen if necessary (of
...
+
+```stan
+(v1, /*...*/, vn) = T;
+```
+
+For example, if `T` is a tuple of type `tuple(int, real, complex)`,
```
bob-carpenter, line 1404: [optional]
I would use lower case `t` or `x` for the tuple here to avoid confusion with `T` being read as a generic type.

src/reference-manual/types.Rmd
```diff
@@ -1386,6 +1386,43 @@ As with other assignments, promotions will happen if necessary (of
...
+real r;
+complex z;
+
+(i, r, z) = T;
+```
+
+Assigns the result of `T.1` to `i`, the result of `T.2` to `r`, and
+the result of `T.3` to `z`.
+
+These unpacking assignments can be nested if the tuple on the right hand
```
bob-carpenter, line 1418: I think the nesting needs an example.  I'm imagining `(i, (r, z)) = x` or something like that?

src/reference-manual/types.Rmd
```diff
@@ -1386,6 +1386,43 @@ As with other assignments, promotions will happen if necessary (of
...
+side contains nested tuples.
+
+The left hand side of an unpacking assignment must contain only
```
bob-carpenter, line 1421: This seems to say it's general, but the earlier description says it's just variables.  Which is it?  And if it's indexed, how do we avoid aliasing or check when it's now a runtime issue?

bob-carpenter (changes requested):
Typo and some questions that should be resolved.

src/reference-manual/types.Rmd
WardBrian: Yes. I've updated the existing example to do both at once following some re-ordering of paragraphs here.

src/reference-manual/types.Rmd
WardBrian: Updated the earlier description. See above for how we avoid aliasing, but the short version is the statement that the same variable may not appear more than once is correct. This is a bit overly-conservative, but does prevent the aliasing problem for indexed assignments appearing here.

bob-carpenter (approved):
Thanks for the updates.

PR #765, opened 2024-04-21
Update doc to use stock RTools toolchain

andrjohns:
#### Submission Checklist

- [x] Builds locally

#### Summary

As of [this Math PR](https://github.com/stan-dev/math/pull/2999), CmdStan and models can build and sample with the stock RTools toolchain (no need to install `mingw32-make` or `gcc` via msys2).

This PR updates the docs to remove all references to `mingw32-make`, and simplifies the windows toolchain setup to simply installing RTools43 and adding the folders to `PATH`

I've also removed the instructions for RTools40 and RTools42, since I don't think we want to continue maintaining instructions for every version of RTools that is released (this is mostly my opinion, happy to revert)

#### Copyright and Licensing

Please list the copyright holder for the work you are submitting (this will be you or your assignee, such as a university or company): Andrew Johnson

### DIFF
--- a/src/cmdstan-guide/installation.qmd
+++ b/src/cmdstan-guide/installation.qmd
@@ -396,7 +396,7 @@ g++ --version
 make --version
 ```
 
-CmdStan is known compatible with the RTools43 toolchain. The toolchais will require
+CmdStan is known compatible with the RTools43 toolchain. The toolchain will require
 updating your `PATH` variable, See [these instructions](https://helpdeskgeek.com/windows-10/add-windows-path-environment-variable/)
 for details on changing the `PATH` if you are unfamiliar. The following instructions will assume that the default installation
 directory was used, so be sure to update the paths accordingly if you have chosen a different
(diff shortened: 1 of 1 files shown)


src/cmdstan-guide/installation.qmd
```diff
@@ -399,93 +387,38 @@ xcode-select --install
...
-pacman -Sy mingw-w64-i686-make            # 32-bit g++-8
-pacman -Sy mingw-w64-x86_64-make          # 64-bit g++-8
-pacman -Sy mingw-w64-ucrt-x86_64-make     # 64-bit g++-10 (UCRT)
-```
-
-##### RTools42 & RTools43
-
-Both RTools42 & RTools43 provide 64-bit UCRT toolchains, where RTools42 provides g++-10 and
-RTools43 provides g++-12. The installation/configuration is identical for both toolchains.
+RTools43 provides a 64-bit UCRT toolchains with GCC 12.3.
```
WardBrian, line 408: I think only describing rtools43 is fine. This is probably a bit too short and technical to be useful to most users, though.

1. Are we sure that the `x86_64-w64-mingw32.static.posix` folder even is UCRT based?
2. If it is, we should probably retain the note that the UCRT requires windows 10+?

WardBrian (changes requested):

src/cmdstan-guide/installation.qmd
andrjohns: Yep it's definitely UCRT, [mentioned in the toolchain doc](https://cran.r-project.org/bin/windows/Rtools/rtools43/rtools.html).

Good point on keeping the Windows 10 note, didn't mean to delete that.

Any thoughts on changes for making it less technical? I'm not sure what would be useful here

WardBrian: The best thing might be to say something like "RTools 43 provides all of the necessary tools to build Stan, including a C++ compiler and the `make` tool" and then link to those docs for more specifics?

andrjohns: Good shout, will do

WardBrian (approved):

PR #785, opened 2024-05-24
warn about the computation cost of GPs

avehtari:
Every now and then someone tries to use GPs with Stan for huge datasets. I think it can save time if realizing Stan GPs are not scaling well.


bob-carpenter (approved):
looks good

PR #788, opened 2024-05-27
Improve docs with respect to statistical terminology

avehtari:
Now that sampling statement has been changed to distribution statement, I continued checking the terms used
- checked that every mention of likelihood refers to likelihood, and replaced with more appropriate terms if referring to data model or distribution
- changed all "sampling distribution" to "data model" or something else, as 1) "sampling" word is heavily overloaded in the doc anyway and in some places it's not clear whether sampling is referring to sampling from the posterior, 2) "sampling distribution" has a frequentist usage which can mislead the reader in Bayesian modeling context
- changed "predictive likelihood" to "posterior predictive distribution"
- couple more distribution statements
- fixed some other sentences and added some clarifications
- fixed some typos

### DIFF
--- a/src/stan-users-guide/cross-validation.qmd
+++ b/src/stan-users-guide/cross-validation.qmd
@@ -190,7 +190,7 @@ of $\tilde{y}$ given the predictors and training data,
 \\[4pt]
 & \approx & \frac{1}{M} \sum_{m = 1}^M \tilde{y}_n^{(m)}
 \end{eqnarray*}
-where $\tilde{y}_n^{(m)}$ is drawn from the sampling distribution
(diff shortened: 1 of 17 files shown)


avehtari:
I added to this PR also changing the section title "Sampling statement" in functions reference, which I had missed earlier

bob-carpenter:
Thanks, @avehtari and @WardBrian --- this is a really great idea.  I never understood before what people meant by "aleatoric" uncertainty.  I describe this in other words later in the guide for posterior predictive inference which has terms p(y' | theta) for the aleatoric uncertainty and p(theta | y) for the epistemic uncertainty.

PR #796, opened 2024-08-04
clarify range indexing, fixes #638

### LINKED ISSUE #638: Clarify index range construction and its difference from R behaviour
#### Summary:

A colleague has mentioned that it would be helpful for the manual to be more explicit about Stan's behaviour with range indexing that involves operations.

For example:

```stan
transformed data {
  array[5] int Y = {1, 2, 3, 4, 5};
  int i = 3;
  int p = 5;

  array[3] int Y_subset = Y[p-i:p-1];  // Y[2:4]
}
```

It would be good to be explicit that the Stan will parse the indexing in `Y[p-i:p-1]` as:
```stan 
(p-i):(p-1)
```
 
As opposed to:
```stan
p - (i:p) - 1
```

Noting that the latter behaviour is how the indexing would be parsed in R:
```r
> i <- 3
> p <- 5
> p-i:p-1
[1]  1  0 -1
```

#### Current Version:
v2.18.0

avehtari:
clarify range indexing, fixes #638

### DIFF
--- a/src/reference-manual/expressions.qmd
+++ b/src/reference-manual/expressions.qmd
@@ -783,8 +783,8 @@ In addition to single integer indexes, as described in
 Multiple indexes can be integer arrays of indexes, lower
 bounds, upper bounds, lower and upper bounds, or simply shorthand for
 all of the indexes.  If the upper bound is smaller than the lower bound, 
-the range is empty (unlike, e.g. in R). Upper bound and lower bound can be 
-expressions that evaluate to integer. A complete table of index types is 
+the range is empty (unlike, e.g., in R). The upper bound and lower bound can be 
+expressions that evaluate to integer. A complete list of index types is 
 given in the following table.
 
 ##### Indexing Options Table {- #index-types-table}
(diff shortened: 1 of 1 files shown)


src/reference-manual/expressions.qmd
```diff
@@ -782,8 +782,10 @@ In addition to single integer indexes, as described in
 [the language indexing section](#language-indexing.section), Stan supports multiple indexing.
 Multiple indexes can be integer arrays of indexes, lower
 bounds, upper bounds, lower and upper bounds, or simply shorthand for
-all of the indexes.  A complete table of index types is given in the
-following table.
+all of the indexes.  If the upper bound is smaller than the lower bound, 
+the range is empty (unlike, e.g. in R). Upper bound and lower bound can be
```
bob-carpenter, line 786: "e.g." always gets a comma after it.  

Upper bound and lower bound -> The upper bound and lower bound

A complete table -> A complete list (rather than using "table" twice)

bob-carpenter (changes requested):
Content's good---I'm just being picky on grammar.

bob-carpenter (approved):
Very clever!  Thanks.

PR #797, opened 2024-08-05
clarify empty array expression

### LINKED ISSUE #621: length 0 array documentation
This issue stems from https://github.com/stan-dev/stanc3/issues/1286 where there was some confusion on length 0 arrays. It becomes apparent that there is ambiguity in the type of the array with length 0. I think we should add documentation around this and state clearly that `{}` is not a valid constructor. 

I believe the relevant section that can be expanded is in the reference manual at https://mc-stan.org/docs/reference-manual/vector-matrix-array-expressions.html#array-expressions.

avehtari:
Fixes #621

The restriction on empty array expression is mentioned later, but it is easy to miss, so clarified it now also earlier with link to the later section. Also added the suggested way to create an empty array.

### DIFF
--- a/src/reference-manual/expressions.qmd
+++ b/src/reference-manual/expressions.qmd
@@ -301,7 +301,7 @@ The empty expression `[ ]` is ambiguous and therefore is not
 allowed and similarly expressions such as `[ [ ] ]` or
 `[ [ ], [ ] ]` are not allowed.
 
-#### Empty vectors and matrices {-}
+### Empty vectors and matrices {-}
 
 If needed, it is possible to create an empty vector with
 ```stan
@@ -347,7 +347,7 @@ array[2, 3] int b = { { 1, 2, 3 },
                 { 4, 5, 6 } };
 ```
 
-#### Empty arrays {-}
+### Empty arrays {-}
 
 The empty array expression (`{ }`) is not allowed. See more about restrictions on array expressions in subsection [Restrictions on values](#array-expressions).
 
--- a/src/reference-manual/expressions.qmd
+++ b/src/reference-manual/expressions.qmd
@@ -334,7 +334,21 @@ array[2, 3] int b = { { 1, 2, 3 },
                 { 4, 5, 6 } };
 ```
 
+#### Empty arrays {-}
+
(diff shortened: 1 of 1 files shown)


src/reference-manual/expressions.qmd
```diff
@@ -417,6 +418,13 @@ zero-element array.
 array[0] int a;   // a is fully defined as zero element array
 ```
 
+If needed, it is also possible to create an empty array with
+```stan
+rep_array(0, 0)
+```
+where the type of the first argument determines the type of the array
+(e.g., real or int).
```
bob-carpenter, line 426: The real and int should be escaped as `real` and `int`. (back ticks)

bob-carpenter: I think it'd be worth pulling out a subsection and highlighting the empty array can be any type, not just a scalar.


### Empty arrays {-}

The empty array expression (`{ }`) is not allowed. See more about restrictions on array expressions in subsection [Restrictions on values](#array-expressions).


If needed, it is also possible to create an empty array with the expression

```stan
rep_array(e, 0)
```
where the first expression `e` determines the type of the array.  For example, `rep_array(0.0, 0)` returns an empty real array of type `real[]`, whereas `rep_array({123}, 0)` returns an empty two dimensional integer array of type `int[ , ]`.  Only the type of the first argument is used, so the integer arrays `{123}` and `{0}` produce equivalent values.
```

bob-carpenter (changes requested):
I suggested a more extensive subsection expanding on what was submitted.

src/reference-manual/expressions.qmd
avehtari: Github shows the markdown in the comment so, that I can't see how many `#` you used for the subsection "Empty arrays". Also I'm not certain where you would like to have that subsection.

avehtari:
Added also instruction how to create empty vectors and matrices

bob-carpenter:
Sorry about that---you can look at the source with the "edit" button.  Three hashes.

The only reason I pulled it out into a section is to make it more noticeable and better indexed by Google (though I'm not sure the indexing is helped much any more by level 3 headers---it used to be).

bob-carpenter (approved):
Looks good.  Thanks.

PR #799, opened 2024-08-05
fix and clarify exp_mod_normal parameter names, fixes #429

### LINKED ISSUE #429: exp_mod_normal parameter lambda is inverse scale, not shape
#### Summary:
In the current Stan Function Reference manual Version 2.28. For the Exponentially modified normal distribution, the parameter lambda is described as being the shape, I think it is actually the inverse scale, or rate parameter. 

#### Description:
Describe the issue as clearly as possible.

#### Additional Information:

I tried to follow the contributing guidelines but could not find a "currently open next docs issue"

#### Current Version:
v2.18.0

avehtari:
fixes #429 

Used Wikipedia naming https://en.wikipedia.org/wiki/Exponentially_modified_Gaussian_distribution, and added one sentence to clarify the role of the different parameters.

### DIFF
--- a/src/functions-reference/unbounded_continuous_distributions.qmd
+++ b/src/functions-reference/unbounded_continuous_distributions.qmd
@@ -380,6 +380,8 @@ and scale `sigma` dropping constant additive terms.
 
 ## Exponentially modified normal distribution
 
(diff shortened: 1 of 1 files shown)


bob-carpenter (approved):
Thanks!

PR #800, opened 2024-08-05
Basic docs for jacobian +=

WardBrian:
#### Submission Checklist

- [x] Builds locally

#### Summary

Docs for https://github.com/stan-dev/stanc3/pull/1435

#### Copyright and Licensing

Simons Foundation

### DIFF
--- a/src/reference-manual/statements.qmd
+++ b/src/reference-manual/statements.qmd
@@ -15,7 +15,7 @@ are used --- if they do not, the behavior is undefined.
 The basis of Stan's execution is the evaluation of a log probability
 function (specifically, a log probability density function) for a given
 set of (real-valued) parameters. Log probability functions can be
-constructed by using distribution statements and log probability increment 
+constructed by using distribution statements and log probability increment
 statements.  Statements may be grouped
 into sequences and into for-each loops.  In addition, Stan allows
 local variables to be declared in blocks and also allows an empty
@@ -248,7 +248,8 @@ of the posterior up to an additive constant.  Data and transformed
 data are fixed before the log density is evaluated.  The total log
 probability is initialized to zero.  Next, any log Jacobian
 adjustments accrued by the variable constraints are added to the log
-density (the Jacobian adjustment may be skipped for optimization).
+density (the Jacobian adjustment may be skipped for maximum likelihood estimation
+via optimization).
 Distribution statements and log probability increment statements may add to the log
 density in the model block.  A log probability increment statement
 directly increments the log density with the value of an expression as
@@ -370,9 +371,9 @@ or in functions ending with `_jacobian` to mimic the log Jacobian
 adjustments accrued by built-in variable transforms.
 
 Similarly to those implemented for the built-in transforms, these Jacobian adjustment
-may be skipped for optimization.
+may be skipped for maximum likelihood estimation via optimization.
 
-For example, here is a program which re-creates the existing
+For example, here is a program which recreates the existing
 [`<upper=x>` transform](transforms.qmd#upper-bounded-scalar) on real numbers:
 
 ```stan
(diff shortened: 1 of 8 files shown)


src/reference-manual/statements.qmd
```diff
@@ -361,6 +361,38 @@ including arrays of vectors and matrices.   For container arguments,
...
+## Increment log density with a change of variables adjustment
+
+
+A variant of the `target +=` statement described above is the
+`jacobian +=` statement. This can be used in the transformed parameters block
+or in functions ending with `_jacobian` to mimic the log Jacobian
+adjustments accrued by built-in variable transforms.
+
+Similarly to those implemented for the built-in transforms, these Jacobian adjustment
+may be skipped for optimization.
```
bob-carpenter, line 373: I would say "may be skipped for maximum likelihood estimation via optimization" to make it clear the optimization isn't speed related here.  If you leave the transforms and optimize, you get the MAP (max a posteriori) estimate.

src/reference-manual/statements.qmd
```diff
@@ -361,6 +361,38 @@ including arrays of vectors and matrices.   For container arguments,
...
+
+For example, here is a program which re-creates the existing
```
bob-carpenter, line 375: "recreate" isn't hyphenated

src/reference-manual/statements.qmd
```diff
@@ -361,6 +361,38 @@ including arrays of vectors and matrices.   For container arguments,
...
+}
+data {
+  real ub;
+}
+parameters {
+  real b_raw;
+}
+transformed parameters {
+  real b = upper_bound_jacobian(b_raw, ub);
+}
```
bob-carpenter, line 393: I think you want to go further and illustrate a positive-constrained distribution on `b`, such as

```stan
model {
  b ~ lognormal(0, 1);
}
```

src/reference-manual/statements.qmd
```diff
@@ -361,6 +361,38 @@ including arrays of vectors and matrices.   For container arguments,
...
+```
+
 ### Accessing the log density {-}
 
 To access accumulated log density up to the current execution point,
```
bob-carpenter, line 398: (I realize this isn't yours, but it should be "the accumulated"

src/stan-users-guide/user-functions.qmd
```diff
@@ -287,6 +287,32 @@ transformed parameters {
...
+data {
+  real ub;
+}
+parameters {
+  real b_raw;
+}
+transformed parameters {
+  real b = upper_bound_jacobian(b_raw, ub);
+}
+```
```
bob-carpenter, line 315: This is a duplicate and I'd like a sampling statement for `b` in the model block.

src/stan-users-guide/using-stanc.qmd
```diff
@@ -647,7 +647,9 @@ Warning:
     Left-hand side of distribution statement (~) may contain a non-linear
     transform of a parameter or local variable. If it does, you need
     to include a target += statement with the log absolute determinant
-    of the Jacobian of the transform.
+    of the Jacobian of the transform. You could also consider defining
```
bob-carpenter, line 650: You could also consider defining -> Alternatively, you could define
using -> use

bob-carpenter (changes requested):
Minor corrections.

src/stan-users-guide/using-stanc.qmd
WardBrian: This is a copy of a warning message stanc3 emits in pedantic mode, so it's wording would need to be changed there

src/reference-manual/statements.qmd
WardBrian: That is no longer "just" a recreation of the built in transform, so I've added it but commented out as an example

src/stan-users-guide/user-functions.qmd
WardBrian: Duplication is intentional here as it's in a very different part of the docs, but I can do something unique if you prefer

src/reference-manual/statements.qmd
bob-carpenter: Good point.  Then it probably makes more sense to add it as a separate comment to say that you can then use a distribution on the transformed parameter.

src/stan-users-guide/user-functions.qmd
bob-carpenter: Understood.  I'd change it up just for variety, but I can see the motivation for duplicating, so up to you.

src/stan-users-guide/using-stanc.qmd
bob-carpenter: Aha.  Our wording's not so great and if we change it, all hell breaks loose on the R side given their goofy unit testing in other packages.

PR #802, opened 2024-08-06
cross-link reverse functions to not miss other supported types

### LINKED ISSUE #798: Supported argument types for reverse?
`reverse` function doc https://mc-stan.org/docs/functions-reference/matrix_operations.html#reverse-functions says the input needs to be a `vector` or `row_vector`, but it seems to work also with `array`

avehtari:
fixes #798 

unified also the section and anchor names

### DIFF
--- a/src/functions-reference/array_operations.qmd
+++ b/src/functions-reference/array_operations.qmd
@@ -545,7 +545,7 @@ Number of components of v less than v[s]
 Number of components of v less than v[s]
 {{< since 2.0 >}}
 
-## Reversing functions {#reversing-functions}
+## Reverse functions {#reverse-functions}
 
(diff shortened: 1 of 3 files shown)


avehtari:
@WardBrian should this be closed as you did #803 ?

WardBrian:
Maybe? I think its worth leaving #798 (possibly with a new title) to note that there could still be a smarter re-organization of the various overloads, but it is true for functions beyond just `reverse` (I myself ran into it last week with `log_sum_exp` living on two different pages)

WardBrian:
I think this can be closed, but it's probably still worth thinking if we should reorganize the functions by what they do rather than what types they work over, which would be the real solution to #798

PR #817, opened 2024-09-24
Edit to macOS C++ install docs

JoFrhwld:
#### Submission Checklist

- [x] Builds locally

#### Summary

As currently written, the C++ toolchain documentation for macOS sends users first to install Xcode from the app store. I've placed a check on whether clang++ and make are installed above that. I've also replaced the instructions to install Xcode from the app store with just the `xcode-select --install` instruction. My reasoning for this is:

##### Adding a check first

- It makes the MacOS instructions more parallel with the Linux and Windows instructions, which also have a checking step before an installation step.
- Some people who are less familiar with these tools may, nevertheless, have installed the Xcode command line tools previously, but don't necessarily remember that they have done so.
- The documented experience of what options are available, and what will happen, will be different depending on whether or not command line tools have already been installed.

##### Removing Xcode app install
The documented steps for installing command line tools through the Xcode app appear to be outdated at least by Xcode v15.4. There is no `Xcode>Preferences` menu, and under `Xcode>Settings`, there is no Downloads tab. In fact, poking around, it's not entirely clear to me _how_ to install command line tools within the Xcode app.

Installing command line tools with `xcode-select --install` still works, and involves fewer steps.

#### Copyright and Licensing

Please list the copyright holder for the work you are submitting (this will be you or your assignee, such as a university or company): Josef Fruehwald

### DIFF
--- a/src/cmdstan-guide/installation.qmd
+++ b/src/cmdstan-guide/installation.qmd
@@ -373,7 +373,7 @@ xcode-select --install
 Select "Install" in the window that opens.
 
 After the installation completes, you can double check that
-istallation was successful by reopening the Terminal and 
+installation was successful by reopening the Terminal and 
 running:
 ```
 clang++ --version
--- a/src/cmdstan-guide/installation.qmd
+++ b/src/cmdstan-guide/installation.qmd
@@ -380,6 +380,9 @@ clang++ --version
 make --version
 ```
 
+You can read more about Xcode on its site: 
+[https://developer.apple.com/xcode/](https://developer.apple.com/xcode/)
+
 We don't recommend trying to use the GNU C++ compiler, available via Homebrew,
 based on the number of reports of installation difficulties from Mac users on GitHub
 as well as the Stan forums.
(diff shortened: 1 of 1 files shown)


WardBrian:
Thank you for updating these! I think it is worth keeping a link to https://developer.apple.com/xcode/ if people want to read more

mitzimorris (changes requested):
~~why delete instructions on how to get Xcode from Apple?~~
nevermind - agree w/ your reasoning and Brian's comment.

one further suggested addition for Mac users:  don't use gnu++ via homebrew:

"We don't recommend trying to use the GNU C++ compiler, available via Homebrew,
based on the number of reports of installation difficulties from Mac users on GitHub
as well as the Stan forums."

bob-carpenter:
Thanks for updating, @JoFrhwld.  I was coming here to add this from Apple:

> You can install the Command Line Tools package by running the xcode-select --install command. Note: macOS comes bundled with xcode-select , a command-line tool that is installed in /usr/bin . It allows you to manage the active developer directory for Xcode and other BSD development tools.

From:  https://developer.apple.com/library/archive/technotes/tn2339/_index.html

JoFrhwld:
I've added both @mitzimorris' admonition against the homebrew version, as well restored a link to the Xcode site as suggested by @WardBrian. Thanks for finding the reference to `xcode-select`'s general availability, @bob-carpenter

mitzimorris (approved):
many thanks for this cleanup!

PR #830, opened 2024-11-13
Update one-dimensional-integrals.qmd

avehtari:
#### Summary

Clarify that `integrate_1d` non-used parameter and data arguments can be any arrays, e.g. `{0}`, if the integrand function does not use those arguments. Creating `{0}` is much easier than creating a zero-length array.

#### Copyright and Licensing

### DIFF
--- a/src/stan-users-guide/one-dimensional-integrals.qmd
+++ b/src/stan-users-guide/one-dimensional-integrals.qmd
@@ -48,13 +48,11 @@ not a function of our parameters.
 The function defining the integrand must have exactly the argument types and
 return type of `normal_density` above, though argument naming is not important.
 Even if `x_r` and `x_i` are unused in the integrand, they must be
-included in the function signature. This may require passing in zero-length arrays
-for data or parameters if the integral does not involve
-data or parameters. Creating zero length arrays is more complicated than creating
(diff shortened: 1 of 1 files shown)


WardBrian:
I think we should remove the word "require" from the prior sentence and just say if they're unused, *any* array can be passed, and the easiest is probably {0}

Note that this will end up allocating some small pieces of memory each call, as opposed to doing it with true 0 length arrays. I'm not sure if we want to get into that level of detail

avehtari:
@WardBrian do you want to make a proposal of a complete sentence(s)?

WardBrian:
Sure, instead of 

> This may require passing in zero-length arrays
for data or a zero-length vector for parameters if the integral does not involve
data or parameters.

Something like

> Even if the integral does not involve some of these, they must still
be supplied _some_ value. The most efficient will be a zero-length array
or vector, which can be created with `rep_array(0, 0)` and `rep_vector(0, 0)`, respectively.
Other options include an uninitialized variable declared with size 0, which is equivalent
to the above, or any easy value, such as `{0}`

avehtari:
I used your version with a minor edit.

WardBrian (approved):

PR #832, opened 2024-11-14
Fix log_mix signature

### LINKED ISSUE #831: Documentation of vectorized log_mix is wrong
#### Summary:
image

[This signature ](https://mc-stan.org/docs/functions-reference/real-valued_basic_functions.html#composed-functions)does not exist. log_mix is not vectorized in the sense of other functions. The 'vectorized'
signature takes a vector of mixing proportions and a vector of log densities 

#### Additional Information:
Math doc here:
https://mc-stan.org/math/prim_2fun_2log__mix_8hpp.html

WardBrian:
#### Submission Checklist

- [x] Builds locally

#### Summary

Closes #831. I'm happy to take further suggestions on the wording here.

#### Copyright and Licensing

Simons Foundation

### DIFF
--- a/src/functions-reference/real-valued_basic_functions.qmd
+++ b/src/functions-reference/real-valued_basic_functions.qmd
@@ -1601,7 +1601,7 @@ The logarithm of one minus the natural exponentiation of x
 Return the natural logarithm of the difference of the natural
 exponentiation of x and the natural exponentiation of y. \begin{equation*}
 \mathrm{log\_diff\_exp}(x,y) = \begin{cases} \log(\exp(x)-\exp(y)) &
-\text{if } +\infty > x \ge y \\[6pt] 
+\text{if } +\infty > x \ge y \\[6pt]
 \textrm{NaN} & \text{otherwise} \end{cases}
 \end{equation*}
 
@@ -1628,11 +1628,16 @@ proportion theta, defined by \begin{eqnarray*}
 \lambda_1, \ \log(1 - \theta) + \lambda_2\right). \end{eqnarray*}
 {{< since 2.6 >}}
 
-<!-- R; log_mix; (T1 theta, T2 lp1, T3 lp2); -->
-\index{{\tt \bfseries log\_mix }!{\tt (T1 theta, T2 lp1, T3 lp2): real}|hyperpage}
+<!-- R; log_mix; (T1 thetas, T2 lps); -->
+\index{{\tt \bfseries log\_mix }!{\tt (T1 thetas, T2 lps): real}|hyperpage}
 
-`R` **`log_mix`**`(T1 theta, T2 lp1, T3 lp2)`<br>\newline
-Vectorized implementation of the `log_mix` function
+`R` **`log_mix`**`(T1 thetas, T2 lps)`<br>\newline
+
+Extension of the two-mixture case above to more densities.
+Return the log mixture of the log densities stored in `lps`,
+where each log density has the mixing proportion given by
+the same index in `thetas`.
+
 {{< since 2.26 >}}
 
 <!-- R; log_sum_exp; (T1 x, T2 y); -->
--- a/src/functions-reference/real-valued_basic_functions.qmd
+++ b/src/functions-reference/real-valued_basic_functions.qmd
@@ -1633,13 +1633,20 @@ proportion theta, defined by \begin{eqnarray*}
 
 `R` **`log_mix`**`(T1 thetas, T2 lps)`<br>\newline
 
-Extension of the two-mixture case above to more densities.
-Return the log mixture of the log densities stored in `lps`,
-where each log density has the mixing proportion given by
-the same index in `thetas`.
+Calculates the log mixture density given `thetas`,
+mixing proportions which should be between 0 and 1 and sum to 1,
+and `lps`, log densities.
+These two containers must have the same length.
 
-In other words, `log_mix(a,b,c)` could be equivalent written
-to use this signature by writing `log_mix({a, 1-a}, {b, c})`.
+\begin{eqnarray*}
+\mathrm{log\_mix}(\theta, \lambda)
+& = & \log \!\left( \sum_{n=1}^N \theta_n * \exp(\lambda_n) \right) \\[3pt]
+& = & \mathrm{log\_sum\_exp}\!\left(\log(\theta) + \lambda\right).
+\end{eqnarray*}
+
+This is a generalization of the above signature of three arguments to
+more than two densities.
+For example, `log_mix(lambda, lp1, lp2) == log_mix({lambda, 1 - lambda}, {lp1, lp2})`.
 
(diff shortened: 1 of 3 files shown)


src/functions-reference/real-valued_basic_functions.qmd
```diff
@@ -1628,11 +1628,19 @@ proportion theta, defined by \begin{eqnarray*}
...
-\index{{\tt \bfseries log\_mix }!{\tt (T1 theta, T2 lp1, T3 lp2): real}|hyperpage}
+<!-- R; log_mix; (T1 thetas, T2 lps); -->
+\index{{\tt \bfseries log\_mix }!{\tt (T1 thetas, T2 lps): real}|hyperpage}
+
+`R` **`log_mix`**`(T1 thetas, T2 lps)`<br>\newline
+
+Extension of the two-mixture case above to more densities.
+Return the log mixture of the log densities stored in `lps`,
+where each log density has the mixing proportion given by
+the same index in `thetas`.
```
bob-carpenter, line 1639: 1.  `thetas` is presumably required to be a simplex, so that should be mentioned.
2. `thetas` and `lps` have to be the same length, so that should be mentioned.
3.  The result should be given explicitly in math, e.g., `log(SUM_n thetas[n] * exp(lps[n]))` or the more efficient but also more opaque `log_sum_exp(log(thetas) + lps)` (it'd be OK to include both if you think just one would be confusing).

src/functions-reference/real-valued_basic_functions.qmd
```diff
@@ -1628,11 +1628,19 @@ proportion theta, defined by \begin{eqnarray*}
...
+
+In other words, `log_mix(a,b,c)` could be equivalent written
```
bob-carpenter, line 1641: equivalent -> equivalently, but I'd just say,

```
For example, `log_mix(lambda, lp1, lp2) == log_mix({lambda, 1 - lambda}, {lp1, lp2})`.

**Review by bob-carpenter — CHANGES_REQUESTED** [review]
Just some minor clarifications.

File: src/functions-reference/real-valued_basic_functions.qmd (hunk 1 above)
Comment (WardBrian) [reply src/functions-reference/real-valued_basic_functions.qmd:1639]: Oddly enough, it doesn't seem that the math library is checking that thetas is a simplex, just that it is bounded between 0 and 1. Possibly worthy of a math issue?

**Comment by bob-carpenter** [comment]
Yes, this should definitely be checked.  I created an issue:

https://github.com/stan-dev/math/issues/3125

**Comment by bob-carpenter** [comment]
I guess we shouldn't say it's checking, just that the first argument must be non-negative values that sum to 1.

**Comment by spinkney** [comment]
Thanks @WardBrian for fixing this, I have been annoyed by this for a long time and was constantly looking up what this does. 

In fact, I have a similar issue with `log_sum_exp` where I wanted it to `vector[n] a = log_sum_exp(vector[n] b, vector[n] c)` across each element, however it doesn't do this type of elementwise sum. Maybe if we had a `log_add_exp` which I've seen in other languages.

File: src/functions-reference/real-valued_basic_functions.qmd (hunk 1 above)
Comment (spinkney) [reply src/functions-reference/real-valued_basic_functions.qmd:1639]: > Extension of the two-mixture case above to more densities. Return the log mixture of the log densities stored in `lps`, where each log density has the mixing proportion given by the same index in `thetas`.

I would prefer this to say something about how this is intended to be looped across each density observation but is generalized to more than two densities. This connects the warning in the mixture modeling section about why we need to loop.

**Review by bob-carpenter — APPROVED** [review]
Thanks for the updates.

**Comment by pmahling** [comment]
Hi is there a place where we can look up types allowed for this function? This example here allowed theta as a vector (size k), and the ps the densities component as an array size N*k, basically using the same weight.  I am curious if this function allows theta to be same N*k array so we can apply different weights to the densities of each data point.   https://github.com/stan-dev/example-models/blob/master/basic_estimators/normal_mixture_k.stan

**Comment by bob-carpenter** [comment]
> is there a place where we can look up types allowed for this function?

That's what the doc update is addressing.

> his example here allowed theta as a vector (size K), and the ps the densities component as an array size N* K, basically using the same weight.

This is not supported according to the documentation Brian wrote above.  It's just `log_mix(ArrayLike, ArrayLike)`, to put it in Python terms, where `ArrayLike` for Stan means a 1D array of reals, `array[] real`, a vector `vector[]`, or a row vector `row_vector[]`.   We could eventually vectorize this to deal with an array of ArrayLike things.

**Comment by bob-carpenter** [comment]
> Do you mean the normal_mixture_k.stan example is not supported?

Yes, that's what I meant because that's what the doc said.  I just tried that example and it compiled just fine.  It also works with an array of different `theta` values.  

@WardBrian---the doc should mention that this can be vectorized in both arguments.  It's not clear from the types, variable names, or example as rendered above.

Yes, I meant we're updating the functions reference doc to reflect what's really going on.  I just misunderstood the existing signatures.  

I'm not sure what you mean by "like R functions".

**Comment by WardBrian** [comment]
> the doc should mention that this can be vectorized in both arguments. It's not clear from the types, variable names, or example as rendered above.

I do not believe it can be. It appears the second argument (`lps`) supports exactly one level of "array-wrapping".

So `log_mix(vector, vector)` is valid, as is `log_mix(vector, array[] vector)`, but the following are both invalid: `log_mix(array[] vector)` or `log_mix(vector, array[,] vector)`

We don't really have any other functions quite like this, so how to describe it exactly might require some new convention

**Comment by bob-carpenter** [comment]
Ah, I forgot to change my variable in my program.   So these both compile:

```stan
log_mix(vector, vector);
log_mix(vector, array[] vector);
```

But these two don't:

```stan
log_mix(array[] vector, vector);
log_mix(array[] vector, array[] vector);
```

I think we should just describe it right in the doc for this function rather than trying to come up with an indirect convention.

pmahling:
If I use below:

log_mix(lambda, poisson_lpmf(y | pe),poisson_lpmf(y| pc))
lambda, y ,pe, pc are all vector of length n.

is this desired   { lambda[1]*f(y[1]|pe[1])+(1- lambda[1])*f(y[1]|pc[1])}* .... { lambda[n]*f(y[n]|pe[n])+(1- lambda[n])*f(y[n]|pc[n])}

bob-carpenter:
`log_mix(lambda, A, B) = log_sum_exp(log(lambda) + A, log1m(lambda) + B)`.

For Stan, `poisson_lpmf(y | pe) = SUM_n poisson_lpmf(y[n] | pe[n])`.

bob-carpenter:
@pmahling:  We have a very active forum which is a better place to ask Stan questions than closed pull requests:  https://discourse.mc-stan.org

I think those would be equivalent, but I'm not 100% sure, so I'd try both to make sure they yield the same answer.  

Note:  you can use `lp[i, 1]` instead of `lp[i][1]`---they should compile to the same C++ code.

Issue #838, opened 2024-11-27
Update Stan users guide chapter on HMMs


charlesm93:
#### Summary:
The section on HMMs (under Time-Series models: https://mc-stan.org/docs/stan-users-guide/time-series.html#hmms.section) in the Stan users guide currently doesn't use Stan's dedicated functions for HMMs (described here: https://mc-stan.org/docs/functions-reference/hidden_markov_models.html).

I'll update the section and point users towards the convenient functions we have.

charlesm93:
I've taken a closer look at the user guide's section on HMMs, and I find that it treats specific class of HMMs, rather than provide general guidance. 

First, the section focuses on the case where the likelihood (observational distribution) is categorical. Then, it considers:
* the case where the hidden states are observed ("supervised parameter estimation"), including naive coding, coding with sufficient statistics, and coding the posterior analytically
* the case where some of the hidden states are observed ("semisupervised estimation")

I'm not sure that the supervised case is particularly interesting, since we can compute the posterior analytically. Its main interest is setting up the semi-supervised example.

If I look at the HMM suite we developed, its use-case seems a bit orthogonal, in that:
* there are no restrictions on the observational model
* none of the hidden states are observed ("unsupervised" case).

So I think it can make sense to keep the existing example and add an example which uses the HMM suite. I'll base myself on the case study @bbbales2 put together after we released the functions (https://mc-stan.org/users/documentation/case-studies/hmm-example.html).

Two organizational questions:
* does it make sense to have the general HMM example first and then the specific example with supervised and semi-supervised learning?
* should we harmonize notation between the examples, and between the user and functions guide?

My answer to both questions is yes. But I'm leaving it up for discussion.

@bob-carpenter

bob-carpenter:
That all sounds good.  But I think it'd also be OK to take out the direct implementation and instead just use the new HMM methods we provide.  I would start with the pre-baked method and only then go on to coding it yourself if you do leave it.

And by all means unify the notation if it's not too much of a bother.

I included semi-supervised because it's so common to see that in natural language processing settings where the hidden states might represent things like a fixed set of parts of speech and the supervision in the form of labeled training data.  I think that's an edge case.  The bigger thing is when there are covariates informing the transition matrix.

Also, I think the next release will be the first with left and right stochastic matrices.  So it'd make sense to use those as we don't talk about them elsewhere in the *User's Guide*.  

Thanks for taking this on!

charlesm93:
[...]

PR #840, opened 2024-12-06
Issue/838 hmm

charlesm93:
#### Submission Checklist

- [x] Builds locally

#### Summary

Addresses issue #838 and updates user-doc on HMMs.

#### Copyright and Licensing

Please list the copyright holder for the work you are submitting (this will be you or your assignee, such as a university or company): Charles Margossian, Simons Foundation

### DIFF
--- a/src/functions-reference/functions_index.qmd
+++ b/src/functions-reference/functions_index.qmd
@@ -294,6 +294,21 @@ pagetitle: Alphabetical Index
  - <div class='index-container'>[distribution statement](unbounded_discrete_distributions.qmd#index-entry-0c7465aa1beceb6e7e303af36b60e2b847fc562a) <span class='detail'>(unbounded_discrete_distributions.html)</span></div>
 
 
+<a id='beta_neg_binomial_cdf' href='#beta_neg_binomial_cdf' class='anchored unlink'>**beta_neg_binomial_cdf**:</a>
+
+ - <div class='index-container'>[`(ints n | reals r, reals alpha, reals beta) : real`](unbounded_discrete_distributions.qmd#index-entry-ecce69045f7901c0a69ccc67aaae5ddda48304b2) <span class='detail'>(unbounded_discrete_distributions.html)</span></div>
+
+
+<a id='beta_neg_binomial_lccdf' href='#beta_neg_binomial_lccdf' class='anchored unlink'>**beta_neg_binomial_lccdf**:</a>
+
+ - <div class='index-container'>[`(ints n | reals r, reals alpha, reals beta) : real`](unbounded_discrete_distributions.qmd#index-entry-50c63a123857b7480722a4aeb3d884c95317cdc3) <span class='detail'>(unbounded_discrete_distributions.html)</span></div>
+
+
+<a id='beta_neg_binomial_lcdf' href='#beta_neg_binomial_lcdf' class='anchored unlink'>**beta_neg_binomial_lcdf**:</a>
+
+ - <div class='index-container'>[`(ints n | reals r, reals alpha, reals beta) : real`](unbounded_discrete_distributions.qmd#index-entry-fbab2aa3f292996672306979423648761d0ab98f) <span class='detail'>(unbounded_discrete_distributions.html)</span></div>
+
+
 <a id='beta_neg_binomial_lpmf' href='#beta_neg_binomial_lpmf' class='anchored unlink'>**beta_neg_binomial_lpmf**:</a>
 
  - <div class='index-container'>[`(ints n | reals r, reals alpha, reals beta) : real`](unbounded_discrete_distributions.qmd#index-entry-9d835cd7ed67218478d70e1db49912bf2d710a1e) <span class='detail'>(unbounded_discrete_distributions.html)</span></div>
@@ -302,6 +317,11 @@ pagetitle: Alphabetical Index
 <a id='beta_neg_binomial_lupmf' href='#beta_neg_binomial_lupmf' class='anchored unlink'>**beta_neg_binomial_lupmf**:</a>
 
  - <div class='index-container'>[`(ints n | reals r, reals alpha, reals beta) : real`](unbounded_discrete_distributions.qmd#index-entry-1bd0b7fec4d1f3e315d4328c1844d15641058615) <span class='detail'>(unbounded_discrete_distributions.html)</span></div>
+
+
+<a id='beta_neg_binomial_rng' href='#beta_neg_binomial_rng' class='anchored unlink'>**beta_neg_binomial_rng**:</a>
+
+ - <div class='index-container'>[`(reals r, reals alpha, reals beta) : R`](unbounded_discrete_distributions.qmd#index-entry-8c97955cf7e95ec459f434cad32f31a963d196cd) <span class='detail'>(unbounded_discrete_distributions.html)</span></div>
 
 
 <a id='beta_proportion' href='#beta_proportion' class='anchored unlink'>**beta_proportion**:</a>
--- a/src/stan-users-guide/time-series.qmd
+++ b/src/stan-users-guide/time-series.qmd
@@ -646,7 +646,7 @@ When each state $x$ takes a value over a discrete and finite set,
 say $\{1, 2, ..., K\}$, we can use Stan's suite of HMM functions to marginalize
 out $x_{1:N}$ and compute $p(y_{1:N} \mid \phi)$.
 We start by defining the conditional observation distribution, stored in a
-$K \times N$ matrix $\mega$ with
+$K \times N$ matrix $\omega$ with
 $$
 \omega_{kn} = p(y_n \mid x_n = k, \phi).
 $$
@@ -654,7 +654,7 @@ Next, we introduce the $K \times K$ transition matrix, $\Gamma$, with
 $$
 \Gamma_{ij} = p(x_n = h \mid x_{n - 1} = i, \phi).
 $$
-Finally, we define the initial state K_vectir $\rho$, with
+Finally, we define the initial state $K$-vector $\rho$, with
 $$
  \rho_k = p(x_0 = k \mid \phi).
 $$
@@ -712,13 +712,13 @@ model {
   mu ~ normal(0, 1);
   sigma ~ normal(0, 1);
 
-  // Increment target by p(y | mu, sigma, Gamma, rho)
+  // Increment target by log p(y | mu, sigma, Gamma, rho)
   target += hmm_marginal(log_omega, Gamma, rho);
 }
 ```
 
 The last function `hmm_marginal` takes in all the ingredients of the HMM,
-and computes the relevant marginal distribution.
+and computes the relevant log marginal distribution.
 Finally, we can recover samples from the posterior distribution of the
 hidden states in generated quantities.
 It is also possible to compute the posterior probbability of each hidden state.
@@ -734,4 +734,4 @@ probabilities of each state, and cannot be used to sample jointly all the
 hidden states. Instead, users should rely on `hmm_latent_rng`.
 
 For more details, since the corresponding case study:
-https://mc-stan.org/users/documentation/case-studies/hmm-example.html.
+[https://mc-stan.org/users/documentation/case-studies/hmm-example.html](https://mc-stan.org/users/documentation/case-studies/hmm-example.html).
--- a/src/stan-users-guide/time-series.qmd
+++ b/src/stan-users-guide/time-series.qmd
@@ -662,18 +662,21 @@ Next, we introduce the $K \times K$ transition matrix, $\Gamma$, with
 $$
 \Gamma_{ij} = p(z_n = j \mid z_{n - 1} = i, \phi).
 $$
+(This is a right-stochastic matrix.)
 Finally, we define the initial state $K$-vector $\rho$, with
 $$
  \rho_k = p(z_0 = k \mid \phi).
 $$
+It is common practice to set $\rho$ to be the stationary distribution of the HMM,
+that is $\rho$ is the first eigenvector of $\Gamma$ and solves $\Gamma \rho = \rho$.
 
-As an example, consider a three-state model, with $K = 1, 2, 3$.
-The observations are normally distributed with
+As an example, consider a three-state model with $K=3$.
+The observations are normally distributed conditional on the HMM states with
 $$
   y_n \sim \text{normal}(\mu_k, \sigma),
 $$
 where $\mu = (1, 5, 9)$ and the standard deviation $\sigma$ is the same
-across all observation.
+across all observations.
 The model is then
 
 ```stan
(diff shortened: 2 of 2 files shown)


bob-carpenter:
I took a look through this and while I'm in favor of minimal examples, this one's a bit too minimal.  I would really like to see the model laid out in more detail than the conditional distributions.  Here are some concrete suggestions:

1.  Use `z` for latent discrete parameters.  We already use `x` for covariates, so it's confusing to use `x`.  I've used `z` elsewhere in the user manual for this in the latent discrete parameter chapter.

2. Define the complete data likelihood $p(y, z | \phi) = p(z \mid \phi) \cdot p(y \mid z, \phi)$ and then the marginalization $p(y \mid \phi)$ that we actually fit.

3.  Mention that $\phi$ is actually several parameters.

4.  I find the constrained version of the transition matrix where you insert zeros makes this first example too challenging.  If you want to discuss that case, I'd suggest another section after this one where you talk about imposing structural zeros.  Otherwise it makes the simple case seem too complicated.  And then you can just define

```stan
array[3] simplex[3] gamma_arr;

matrix[3, 3] gamma;
for (n in 1:3) gamma[n] = gamma_arr[n];
```

5.  For the doc, those `mu` and `sigma` are not just the measurement model---that's all the error terms.

6.  Wherever you have repetition, use loops.  It's less error prone and more clear that it's a homogeneous operation:

```stan
for (n in 1:N) {
  for (k in 1:3) {
    log_omega[k, n] = normal_lpdf(y[n] | mu[k], sigma);
  }
}
```

7. Given that you're tying the parameter `sigma` across outputs, you need to mention that.  I'd recommend just keeping this simple with a vector of `sigma` values.

8.  You can't just say "computes the relevant log marginal distribution"---you have to say what that is.  I don't mean including the marginalization algorithm, I just mean as I wrote it above.

9.  For more details, since -> For more details, see.   Also, I wouldn't say "corresponding case study", I'd just say it's a case study on HMMs.  And then you should put it in the bibtex file and cite it properly with reference to Ben (the author).  And you should cite that you "borrowed" the example from Ben Bales's case study.

charlesm93:
I'm rewriting the code to make the transition matrix less constrained (per comment 4) and I wanted to check: we don't have a stochastic matrix type, right?

WardBrian:
We have row and column but not double yet https://mc-stan.org/docs/reference-manual/types.html#stochastic-matrices

charlesm93:
@bob-carpenter I implemented your feedback.

Some questions:

5. What exactly is the difference between a measurement and an error model? I'm ok to use either term, and I know they have different conceptual implication, but I'm wondering if they have a formal definition.

7. I'm keeping `sigma` a scalar. I'm not sure it's simpler conceptually to pass it as a vector.

bob-carpenter:
For (5), I think the idea's that there are three sources of error:  measurement error, modeling error, and sampling error.  For example, sampling error arises when you subsample a population and use that for estimation.  You get modeling error if you use a linear regression for a relationship that's not linear or use normal errors when the errors are skewed, and so on.  If you're weighing things with a scale and you know the scale's biased to the high side, you can correct that measurement error.  You can explicitly add a measurement error model if you know your measurement model (e.g., gravitational lensing is part of the measurement error model; your work with Bruno et al. on deconvolving galactic dust is part of the measurement model for the CMB, etc.).

src/functions-reference/functions_index.qmd
```diff
@@ -294,6 +294,21 @@ pagetitle: Alphabetical Index
  - <div class='index-container'>[distribution statement](unbounded_discrete_distributions.qmd#index-entry-0c7465aa1beceb6e7e303af36b60e2b847fc562a) <span class='detail'>(unbounded_discrete_distributions.html)</span></div>
 
 
+<a id='beta_neg_binomial_cdf' href='#beta_neg_binomial_cdf' class='anchored unlink'>**beta_neg_binomial_cdf**:</a>
```
bob-carpenter, line 297: Why is this PR touching negative binomial?  Are you up to date with the main branch?

src/stan-users-guide/time-series.qmd
```diff
@@ -634,334 +634,116 @@ time of the original iterations.
...
 $$
-The output $y_t$ at time $t$ is generated conditionally independently
-based on the latent state $z_t$.
-
-This section describes HMMs with a simple categorical model for
-outputs $y_t \in \{ 1, \dotsc, V \}$.  The categorical distribution for
-latent state $k$ is parameterized by a $V$-simplex $\phi_k$.  The
-observed output $y_t$ at time $t$ is generated based on the hidden
-state indicator $z_t$ at time $t$,
+When $z_{1:N}$ is continuous, the user can explicitly encode these distributions
```
bob-carpenter, line 647: Isn't `z_{1:N}` a subset of $\\{ 1, ..., K \\}^N$ when $K$ is the number of states?  I don't think it can be continuous.  

Did you perhaps mean to say that the output $y_1, \ldots, y_T$ can be continuous?  It can also be mixed discrete and continuous, it can be multivariate, etc.---it can really be anything.

src/stan-users-guide/time-series.qmd
```diff
@@ -634,334 +634,116 @@ time of the original iterations.
...
+$$
+We start by defining the conditional observation distribution, stored in a
+$K \times N$ matrix $\omega$ with
+$$
+\omega_{kn} = p(y_n \mid z_n = k, \phi).
+$$
+Next, we introduce the $K \times K$ transition matrix, $\Gamma$, with
+$$
+\Gamma_{ij} = p(z_n = j \mid z_{n - 1} = i, \phi).
+$$
```
bob-carpenter, line 664: I think you need to mention that this is a right-stochastic matrix, where we now have that data type built in.

src/stan-users-guide/time-series.qmd
```diff
@@ -634,334 +634,116 @@ time of the original iterations.
...
+Finally, we define the initial state $K$-vector $\rho$, with
+$$
+ \rho_k = p(z_0 = k \mid \phi).
```
bob-carpenter, line 667: I would mention at this point that it is common to take $\rho$ to be the stationary distribution of the HMM.  Many package like movementHMM in R let you set it to the stationary distribution by default.  You can mention that the stationary distribution is the first eigenvector (that's why they called Google's PageRank algorithm, which computed the stationary distribution of their random web walking model using an iterative algorithm, the "billion-dollar eigenvector").  It's the solution for $\pi$ in $\Gamma \cdot \pi = \pi$ for left-stochastic transition matrix $\Gamma$.

src/stan-users-guide/time-series.qmd
```diff
@@ -634,334 +634,116 @@ time of the original iterations.
...
-component indicators form a latent Markov chain.
-
-
-
-### Supervised parameter estimation {-}
 
-In the situation where the hidden states are known, the following
-naive model can be used to fit the parameters $\theta$ and $\phi$.
+As an example, consider a three-state model, with $K = 1, 2, 3$.
+The observations are normally distributed with
```
bob-carpenter, line 671: `K` is an integer, so I think that should just be $K = 3$.

bob-carpenter, line 671: normally distributed conditioned on the HMM state

src/stan-users-guide/time-series.qmd
```diff
@@ -634,334 +634,116 @@ time of the original iterations.
...
+$$
+  y_n \sim \text{normal}(\mu_k, \sigma),
+$$
+where $\mu = (1, 5, 9)$ and the standard deviation $\sigma$ is the same
+across all observation.
```
bob-carpenter, line 676: observation -> observations

src/stan-users-guide/time-series.qmd
```diff
@@ -634,334 +634,116 @@ time of the original iterations.
...
-sentence and is prohibited elsewhere.
-
-### Calculating sufficient statistics {-}
-
-The naive HMM estimation model presented above can be sped up
-dramatically by replacing the loops over categorical distributions
-with a single multinomial distribution.
+parameters {
+  // Rows of the transition matrix
+  array[3] simplex[3] gamma_arr;
```
bob-carpenter, line 687: Ack.  Does this mean you require an array of simplexes, but our `row_stochastic_matrix` type is a matrix so it can't be used.  If so, that's a real bummer, because nobody will ever use the `row_stochastic_matrix` type!  I guess we didn't think the data types through with @SteveBronder.

src/stan-users-guide/time-series.qmd
```diff
@@ -634,334 +634,116 @@ time of the original iterations.
...
 
-The data are declared as before.  The transformed data block
-computes the sufficient statistics for estimating the transition and
-emission matrices.
+  // Initial state
+  simplex[3] rho;
```
bob-carpenter, line 690: Why are there no more priors on the simplexes?  You should at least mention here that they're getting default uniform priors over the simplexes.

src/stan-users-guide/time-series.qmd
```diff
@@ -634,334 +634,116 @@ time of the original iterations.
...
-  // ...
+  // Increment target by log p(y | mu, sigma, Gamma, rho)
+  target += hmm_marginal(log_omega, gamma, rho);
 }
 ```
 
-The model for the supervised data does not change; the unsupervised
-data are handled with the following Stan implementation of the forward
-algorithm.
+The last function `hmm_marginal` takes in all the ingredients of the HMM,
```
bob-carpenter, line 721: Eliminate the comma---English doesn't use commas between conjunctions unless there are more than two.  So it's just "A and B", but it's either "A, B, and C" (Oxford style) or "A, B and C" (defective American style).

src/stan-users-guide/time-series.qmd
```diff
@@ -634,334 +634,116 @@ time of the original iterations.
...
-The Viterbi algorithm can be coded in Stan in the generated quantities
-block as follows.  The predictions here is the most likely state
-sequence `y_star[1], ..., y_star[T_unsup]` underlying the
-array of observations `u[1], ..., u[T_unsup]`.  Because this
-sequence is determined from the transition probabilities
-`theta` and emission probabilities `phi`, it may be
-different from sample to sample in the posterior.
+To obtain samples from the posterior distribution of $z$,
+we use the generated quantities block and draw, for each sample $\phi$,
+a sample from $p(z \mid y, \phi)$.
```
bob-carpenter, line 726: You don't typically need draws from $z$.  If you do need statistics involving $z$, it's often easier to compute them in expectation (i.e., take advantage of the Rao-Blackwellization you get from marginalizing).  For example, if there are low probability but non-zero $z$, they're very hard to evaluate with sampling.  There are examples in the latent discrete parameters chapter.

You can also use the forward-backward algorithm to work out marginally $p(z_n \mid y)$ and you can use Viterbi with backtracking to find the most probable z given y.

src/stan-users-guide/time-series.qmd
```diff
@@ -634,334 +634,116 @@ time of the original iterations.
...
+In effect, MCMC produces draws from $p(\phi \mid y)$
+and with the draw in generated quantities, we obtain draws from
+$p(\phi \mid y) p(z \mid y, \phi) = p(z, \phi \mid y)$.
+It is also possible to compute the posterior probbability of each hidden state,
+that is $\text{Pr}(z_n = k \mid \phi, y)$.
```
bob-carpenter, line 731: Typically this `phi` is marginalized out and we look at  $p(z_n = k | y)$.

src/stan-users-guide/time-series.qmd
```diff
@@ -634,334 +634,116 @@ time of the original iterations.
...
-That is, the model is fit independently of `u`, then those
-parameters used to find the most likely state to have generated
-`u`.
-
+`hmm_hidden_state_prob` returns the marginal
+probabilities of each state, $\text{Pr}(z_n = k \mid \phi, y)$.
+This function cannot be used to compute the joint probability
+$\text{Pr}(z \mid \phi, y)$, because such calculation requires accounting
+for the posterior correlation between the different components of $z$.
+Therefore, `hidden_probs` should NOT be used to obtain posterior samples.
```
bob-carpenter, line 744: NOT -> *not* (that'll make it italics, which is the standard way to add emphasis to typeset text).

This point is a bit more subtle, though.  You can use this to obtain posterior draws of the marginals $z_n \sim p(z_n \mid y)$, but if you put a sequence of draws from $p(z_1 \mid y), \ldots, p(z_N \mid y)$, you don't get a draw of $p(z \mid y)$.

src/stan-users-guide/time-series.qmd
```diff
@@ -634,334 +634,116 @@ time of the original iterations.
...
+Instead, users should rely on `hmm_latent_rng`.
```
bob-carpenter, line 745: Although not necessary, it's nice to have examples of how to do that.  Like:

```stan
generated quantities {
   array[N] int<lower=1, upper=K> z = hmm_latent_rng(...fill-in params here to match example...);
}
```

I know this feels obvious, but it can be hard for the users to put the arguments and result types together.

bob-carpenter (changes requested):
There's a ton of little things to fix, but nothing major.

src/functions-reference/functions_index.qmd
charlesm93: hmmm... that's odd. Let me check.

src/stan-users-guide/time-series.qmd
charlesm93: `z_{1:N}` denotes the latent variables. In general, the domain of `z` can be continuous, say $\mathbb R$, or discrete  $\{1, \cdots, K\}^N$.

src/stan-users-guide/time-series.qmd
charlesm93: We wrote the HMM suites before we had stochastic matrices. We can do a PR and change the arguments.

src/stan-users-guide/time-series.qmd
charlesm93: Ok. Although I was under the impression they didn't use the Oxford comma in the UK. I remember noticing this during StanCon in Cambridge and I figured out the Cambridge folks were just taking a piss. But when I asked some of the locals, it turned out they were familiar with the Oxford comma.

src/stan-users-guide/time-series.qmd
charlesm93: I added "If we desire samples from the posterior distribution of $z$..."

src/stan-users-guide/time-series.qmd
charlesm93: You get $p(z_n = k \mid y)$ by averaging all the MCMC draws. I'll specify this in the doc.

src/functions-reference/functions_index.qmd
charlesm93: As far as I can tell, I am up to date with the main branch, so I'm not sure how this error occurred. I suppose I could simply delete the extra lines.

charlesm93:
@bob-carpenter I completely missed your review. Sorry about that. Hopefully we'll merge this in soon.

PR #858, opened 2025-01-28
Update GP ARD docs to use built-in cov function

### LINKED ISSUE #856: Should User's Guide page on Gaussian Process ARD use the built-in covariance function?
In the [Automatic Relevance Determination](https://mc-stan.org/docs/stan-users-guide/gaussian-processes.html#automatic-relevance-determination) section of the User's Guide, there is this passage:

> The implementation of automatic relevance determination in Stan is straightforward, though it currently requires the user to directly code the covariance matrix. We’ll write a function to generate the Cholesky of the covariance matrix called `L_gp_exp_quad_cov_ARD`.

followed by a custom implementation in the example. In the Functions Reference page for Gaussian Process Covariance Functions under the section for the [Exponentiated quadratic kernel](https://mc-stan.org/docs/functions-reference/matrix_operations.html#exponentiated-quadratic-kernel), there is a function signature for 

```matrix gp_exp_quad_cov(vectors x, real sigma, array[] real length_scale)```

which has an implementation of varying length scales across input space dimensions. To my eye, it looks like you could accomplish the custom `L_gp_exp_quad_cov_ARD` with the same code in the Latent variable GP example:

```stan

matrix[N, N] L_K;
matrix[N, N] K = gp_exp_quad_cov(x, alpha, rho);
for (n in 1:N) {
  K[n, n] = K[n, n] + delta;
}
L_K = cholesky_decompose(K);
```

where here `rho` is an array of reals. 

Not sure if I'm missing something here? Happy to told I'm missing something. In my brief test case, the custom ARD function and the built in code above gave similar results (although the built-in was much faster!).

amas0:
#### Submission Checklist

- [x] Builds locally

#### Summary

Slight rewrite of the Automatic Relevance Determination section the Stan User's guide on Gaussian processes. 

Proposed to fix #856

### DIFF
--- a/src/stan-users-guide/gaussian-processes.qmd
+++ b/src/stan-users-guide/gaussian-processes.qmd
@@ -547,33 +547,11 @@ of $x_1$ is higher [@RasmussenWilliams:2006, page 80].
 The collection of $\rho_d$ (or $1/\rho_d$) parameters can also be
 modeled hierarchically.
 
-The implementation of automatic relevance determination in Stan is
-straightforward, though it currently requires the user to directly code the
-covariance matrix. We'll write a function to generate the Cholesky of the
-covariance matrix called `L_gp_exp_quad_cov_ARD`.
+The implementation of automatic relevance determination is a straightforward
+extension of the one-dimensional case by modifying `rho` to be an array.
 
 
 ```stan
-functions {
-  matrix L_gp_exp_quad_cov_ARD(array[] vector x,
-                            real alpha,
-                            vector rho,
-                            real delta) {
-    int N = size(x);
-    matrix[N, N] K;
-    real sq_alpha = square(alpha);
-    for (i in 1:(N-1)) {
-      K[i, i] = sq_alpha + delta;
-      for (j in (i + 1):N) {
-        K[i, j] = sq_alpha
-                      * exp(-0.5 * dot_self((x[i] - x[j]) ./ rho));
-        K[j, i] = K[i, j];
-      }
-    }
-    K[N, N] = sq_alpha + delta;
-    return cholesky_decompose(K);
-  }
-}
 data {
   int<lower=1> N;
   int<lower=1> D;
@@ -584,7 +562,7 @@ transformed data {
   real delta = 1e-9;
 }
 parameters {
-  vector<lower=0>[D] rho;
+  array[D] real<lower=0> rho;
   real<lower=0> alpha;
   real<lower=0> sigma;
   vector[N] eta;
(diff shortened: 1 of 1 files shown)


bob-carpenter:
Thanks for opening a PR.

I'm sorry I wasn't clear before.  

We try very hard in the *User's Guide* to present fairly optimal code.  If we use less efficient code, our users will copy it and be unhappy.

It is OK to provide a less efficient alternative for pedagogical purposes before introducing the more efficient variant.  Here, you can leave the Cholesky-based version, then note that various pieces add up to the same thing as doing it directly with a dense covariance matrix rather than its Cholesky factor.

bob-carpenter (changes requested):
My last comment was in the wrong place.  Sorry for being misleading before, but we do not want to remove the efficient version of this with Cholesky factors. 

What I would recommend is adding a note that shows the equivalent dense covariance matrix parameterization and mentions it is less efficient than working with Cholesky factors.

We do not want to provide code that users copy that isn't efficient.

amas0:
I apologize, but I seem to have gotten confused along the way here. I
raised this issue because for my problem the current code in the guide is
significantly *less efficient.* The suggested change still uses Cholesky
factors but replaces a manual construction of `K` with the built-in
`gp_exp_quad_cov`. Let me try to clarify a bit:

I had been using the Cholesky factor version of the ARD covariance matrix
that was provided in the User's Guide, i.e. I just copy-pasted the
user-defined `L_gp_exp_quad_cov_ARD`. The wording in the User's guide:
"...though it currently requires the user to directly code the covariance
matrix" implied to me that this code was provided because the approach used
in other sections of the guide (such as under "Latent variable GP") --
where `K` is constructed with `gp_exp_quad_cov` and then the Cholesky
factor `L_K` is constructed with `cholesky_decompose(K)` -- did not work
with varying length scales across the input dimensions.

When I was looking through the Function's reference, I found that
`gp_exp_quad_cov` *did* support varying length scales. So I tried passing
an array of length scales to `gp_exp_quad_cov` and computing the Cholesky
factor from that, instead of using `L_gp_exp_quad_cov_ARD`. When I did
this, the wall time of sampling from my model was significantly faster and
produced similar results.

I don't know what the differences are under the hood, but for my problem
using `L_gp_exp_quad_cov_ARD` seemed to be noticeably less efficient -- my
best guess is that the provided code in the guide manually constructs the
dense matrix `K` instead of doing it with the built-in?

bob-carpenter:
Thanks, @amas0.  My bad.  I didn't see that the `L_...` form wasn't a built-in but was something the user built from scratch with their own squared exponential kernel.

I'll go back and approve now that I see what's going on!

PR #934, opened 2026-03-12
entry for wiener diffusion model

Franzi2114:
#### Summary
This PR adds a user's guide entry for the wiener diffusion model. It is linked to the PRs for the seven parameter diffusion model PDF and CDF functions 
[https://github.com/stan-dev/math/pull/2822](https://github.com/stan-dev/math/pull/2822)
[https://github.com/stan-dev/math/pull/3042](https://github.com/stan-dev/math/pull/3042)

#### Copyright and Licensing

Franziska Henrich


bob-carpenter:
Thanks, so much, @Franzi2114.  If nobody else can get to this, I can review it some time over the next week.

Franzi2114:
Perfec,t thanks @bob-carpenter! I was also in contact  with @WardBrian before I did the PR.

There are still three open points that I didn't manage to incorporate:

- The citations did not work locally although I added the entries in the bib file
- I would prefer to include the .pdf version instead of the .jpeg version of the image as it has much higher quality
- and there could be inserted a link to another subchapter of the User’s guide.

Please let me know when there occur more points that I should change.

WardBrian:
Hi @Franzi2114 --

I've just pushed a commit that addresses each of your comments, since they're mostly quarto-wrangling:

- The citations issue seems to have just been a small typo.
- Quarto supports [multi-format figures](https://quarto.org/docs/authoring/figures.html#multiformat-figures), so as it is currently written the website will use the JPG and the pdf manual will use the PDF. If you have a higher-quality raster image (png, nicer jpg, etc) we can use that on the web version instead
- I added the link and made sure that this new chapter shows up on the sidebar.

WardBrian:
Pinging @bob-carpenter again -- I think the notifications for this got lost while you were away

bob-carpenter:
Thanks for pinging.  I will review this week.

src/bibtex/all.bib
```diff
@@ -1352,3 +1352,63 @@ @article{Henrich2024
   url={https://doi.org/10.3758/s13428-023-02179-1}
 }
 
+@article{Henrich2026,
+  author={Henrich, Franziska and Klauer, Karl Christoph},
+  title={Modeling truncated and censored data with the diffusion model in Stan},
```
bob-carpenter: Capitals in titles need to be escaped, so 

`Stan` -> `{S}tan`

in the title.

src/bibtex/all.bib
```diff
@@ -1352,3 +1352,63 @@ @article{Henrich2024
...
+  number={42},
+  doi={10.3758/s13428-025-02822-z},
+  url={https://doi.org/10.3758/s13428-025-02822-z}
+}
+
+
+@article{Wagenmakers2009,
+ author = {Wagenmakers, Eric-Jan},
+ year = {2009},
+ title = {{Methodological and empirical developments for the Ratcliff diffusion model of response times and accuracy}},
```
bob-carpenter: Only escape the unusual caps, so that BibTeX can style.  This should not have the outer braces and just curly braces around the `R` in `Ratcliff`.

src/bibtex/all.bib
```diff
@@ -1352,3 +1352,63 @@ @article{Henrich2024
...
+ pages = {641--671},
+ volume = {21},
+ number = {5},
+ issn = {0954-1446},
+ journal = {{European Journal of Cognitive Psychology}},
```
bob-carpenter: No double braces.  BibTeX styles sort out which style of capitalization and we only need to provide hints about things that should *always* be capitalized by wrapping in curly braces.

src/bibtex/all.bib
```diff
@@ -1352,3 +1352,63 @@ @article{Henrich2024
...
+ doi = {10.1080/09541440802205067}
+}
+
+
+@article{Ulrich1994,
+ author = {Ulrich, Rolf and Miller, Jeff},
+ year = {1994},
+ title = {{Effects of truncation on reaction time analysis}},
```
bob-carpenter: No double braces.

src/bibtex/all.bib
```diff
@@ -1352,3 +1352,63 @@ @article{Henrich2024
...
+ pages = {34--80},
+ volume = {123},
+ number = {1},
+ journal = {{Journal of Experimental Psychology: General}}
```
bob-carpenter: No double braces.

src/bibtex/all.bib
```diff
@@ -1352,3 +1352,63 @@ @article{Henrich2024
...
+}
+
+
+@article{Ratcliff1998,
+ author = {Ratcliff, Roger and Rouder, Jeffrey N.},
+ year = {1998},
+ title = {{Modelling response times for two-choice decisions}},
```
bob-carpenter: OK, I'm not going to keep writing this, but all the double braces should be removed.

src/bibtex/all.bib
```diff
@@ -1352,3 +1352,63 @@ @article{Henrich2024
...
+@article{Ratcliff1978,
+ author = {Ratcliff, Roger},
+ year = {1978},
+ title = {{A Theory of Memory Retrieval}},
+ pages = {59--108},
+ volume = {85},
+ number = {2},
+ journal = {{Psychological Review}}
+}
+
```
bob-carpenter, line 1543: I can't figure out where else to leave this comment, so I'm putting it here.  

If the image was taken from a publication, then the source needs to be cited, the license of the source, and copyright holder of the source.  Even if that copyright holder is you.

bob-carpenter, line 1543: Also w.r.t. the image, is the .pdf version being used?  The web site will presumably use the .jpg version.  Does the LaTeX automatically grab the .pdf somehow?  If not, remove the .pdf if it's not being used.

src/stan-users-guide/wiener_diffusion_model.qmd
```diff
@@ -0,0 +1,337 @@
+---
+pagetitle: Wiener Diffusion Model
+format:
+  html:
+    default-image-extension: jpg
+---
+
+# Wiener Diffusion Model
+
+Diffusion models, sometimes also called Wiener diffusion models, are among the most frequently used model families in modeling two-alternative forced-choice tasks (see @Wagenmakers2009, for a review). Diffusion models allow to model response times and responses jointly. The basic version of a diffusion model comprises four parameters: the boundary separation, $a$, the relative starting point, $w$, the drift rate, $v$, and the non-decision time, $t0$ [@Ratcliff1978]. In the seven-parameter extension of the diffusion model there are the following three parameters added: the inter-trial variability in relative starting point, $s_w$, the inter-trial variability in drift rate, $s_v$, and the inter-trial variability in non-decision time, $s_{t0}$ [@Ratcliff1998].
```
bob-carpenter: If you could keep lines to <= 80 characters or so, it really helps with git, which is line based.  Otherwise, as you will see below, comments are for paragraphs and you need to fish them out

bob-carpenter: I think we should also cite the last chapter of this book (Chapter 18) as it presents these models from a Stan perpsective:

```
@book{nicenboim2025introduction,
  title={Introduction to {B}ayesian data analysis for cognitive science},
  author={Nicenboim, Bruno and Schad, Daniel J and Vasishth, Shravan},
  year={2025},
  publisher={CRC Press}
}
```

I escaped the "B" for you :-).

src/stan-users-guide/wiener_diffusion_model.qmd
```diff
@@ -0,0 +1,337 @@
...
+
+As a diffusion model describes the decision process for a decision with exactly two choices, there exist reaction time distributions for each response alternative. This means that the probability distribution function ($\text{PDF}$) splits into one part for one response alternative and one part for the other response alternative. In the following, we will refer to one alternative as the _upper response boundary_ and to the other alternative as the _lower response boundary_. The $\text{PDF}$ of the lower response boundary can be obtained when inserting $-v$ and $1-w$ to the $\text{PDF}$ of the upper response boundary. Let's call the $\text{PDF}$ for the lower response boundary $\text{PDF}_0$ and the $\text{PDF}$ for the upper response boundary $\text{PDF}_1$. Then:
+
+$$
+\text{PDF}_0(a,t0,v,w,sv,sw,st0) = \text{PDF}_1(a,t0,-v,1-w,sv,sw,st0)
+$$
+
+Usually, a $\text{PDF}$ integrates to 1. In the case of the diffusion model, only the sum of both parts, $\text{PDF}_0$ and $\text{PDF}_1$, integrates to 1. This is called a _defective_.
+
```
bob-carpenter, line 54: "defective" is just an adjective, so the final sentence still needs a noun.  Did you mean "defective model"?

You could also say "this is called defective" without the determiner "a".

src/stan-users-guide/wiener_diffusion_model.qmd
bob-carpenter: Does "variability" mean "variance" here?  Or is it the scale (standard deviation) of variation?  I think you should be more specific here if possible.

src/stan-users-guide/wiener_diffusion_model.qmd
```diff
@@ -0,0 +1,337 @@
...
+::: {#fig-DDM}
+![](img/Figure_DDM)
+
+Figure 1: Realization of a Four-Parameter Diffusion Process Modeling the Binary Decision Process.
+ **Note.** The parameters are the boundary separation a for two response alternatives, the relative starting point w, the drift rate v, and the non-decision time t0. The decision process is illustrated as a jagged line between the two boundaries. The predicted distributions of the reaction times are depicted as curved lines below and above the response boundaries (blue).
+:::
+
+
+The Stan function `wiener_lpdf()` returns the logarithm of the first-passage time density function for a diffusion model with up to seven parameters for upper boundary responses, $\log(\text{PDF}_1)$. Any combination of fixed and estimated parameters can be specified. In other words, with this implementation it is not only possible to estimate parameters of the full seven-parameter model, but also to estimate restricted models such as the basic four-parameter model, or a five or six-parameter model, or even a one-parameter model when fixing the other six parameters.
```
bob-carpenter: You haven't said what the "first-passage time density function" is.  I'm actually confused here as you've introduced two unnormalized densities, one for upper crossing time and one for lower crossing time.  Also, what is the constraint on $y$?  Is it positive and continuous and how do we tell if it's the positive or negative decision (or however they're conventionally designated if not positive/negative)?

src/stan-users-guide/wiener_diffusion_model.qmd
```diff
@@ -0,0 +1,337 @@
...
+
+For example, it is possible to permit variability in just one or two parameters and to fix the other variabilities to 0, or even to estimate a three-parameter model when fixing more parameters (e.g., fixing the relative starting point at 0.5).
```
bob-carpenter: Will the model still work with variances (or whatever they are) set to zero?  That usually leads to a log(0) or a divide-by-zero in the densities.

src/stan-users-guide/wiener_diffusion_model.qmd
```diff
@@ -0,0 +1,337 @@
...
+ v ∼ normal(2, 3);
+ t0 ∼ normal(0.435, 0.12);
+
+ sv ∼ normal(1, 3);
+ st0 ∼ normal(0.183, 0.09);
+ sw ∼ beta(1, 3);
+
+ // diffusion model
+ for (i in 1:N) {
+  if (resp[i] == 1) { // upper boundary
```
bob-carpenter: The code conflicts with the statement that $y \sim \textrm{wiener\_lpdf}(...)$, because the data is really two things, a reaction time and a response.  Based on the response, the distribution of the reaction time changes.  You already set this up with the complementary times and defectiveness, but the earlier statement in math needs to be amended to match the code.  (It's not just you---I often find myself having to read code to figure out what someone's math means, including my own from the past.)

src/stan-users-guide/wiener_diffusion_model.qmd
```diff
@@ -0,0 +1,337 @@
...
+   target += wiener_full_lpdf(rt[i] | a, t0, w, v, sv, sw, st0);
+  } else { // lower boundary(mirror drift and starting point )
+   target += wiener_full_lpdf(rt[i] | a, t0, 1-w, -v, sv, sw, st0);
+  }
```
bob-carpenter, line 154: I don't know if `wiener\_full\_lpdf` is vectorized, but if it is, it's going to be more efficient  to sort the responses and use a vectorized call.  You don't need to do this here, but it's often useful to say this.  

The thing to do is 

```stan
transformed data {
  real pos_rt[sum(resp)];
  real neg_rt[N - sum(resp)];
  int pos_pos = 1, neg_pos = 1;
  for (n in 1:N) {
    if (resp[n] == 1) {
      pos_rt[pos_pos] = rt[n];
      pos_pos += 1;
    } else {
      neg_rt[neg_pos] = rt[n];
      neg_pos += 1;
    }
  } 
}
```

And then call the relevant likelihoods just once

```stan
pos_rt ~ wiener_full_lpdf(a, t0, w, v, sv, sw, st0);
neg_rt ~ wiener_full_lpdf(a, t0, 1-w, -v, sv, sw, st0)
```

Using the sampling form will automatically drop constants.  Otherwise, I'd suggest using the `lupdf` forms as you won't need the normalization here.

src/stan-users-guide/wiener_diffusion_model.qmd
```diff
@@ -0,0 +1,337 @@
...
+ }
```
bob-carpenter: The way you have it, it's more efficient otherwise pull out repeated constants, though it's probably not going to matter much here given how expensive the density calls are.  That'd be local variables:

```stan
real one_minus_w = 1 - w;
real neg_v = -v;
```

You can either just ignore this, change the code, or write a note mentioning it'd be more efficient to define these once and reuse.

src/stan-users-guide/wiener_diffusion_model.qmd
```diff
@@ -0,0 +1,337 @@
...
+}
+```
+
+### The Data Block
+
+The data should consist of at least three variables:
+
+1. The number of trials N,
+2. the response, coded as 0 = “lower bound” and 1 = “upper bound", and
+3. the reaction times in seconds (not milliseconds).
```
bob-carpenter, line 164: [comment]
I'd have been very tempted to code this as negative reaction times for lower bounds and positive for upper bounds and then that could be built into the density.  I'm guessing this isn't conventional, though, so probably best not to do it here.

src/stan-users-guide/wiener_diffusion_model.qmd
```diff
@@ -0,0 +1,337 @@
...
+
+Note that two different ways of coding responses are commonly used: First, in _response coding_, the boundaries correspond to the two response alternatives. Second, in _accuracy coding_, the boundaries correspond to correct (upper bound) and wrong (lower bound) responses.
+
```
bob-carpenter: Was 1--3 the way to do it for response coding?  If so, it'd be really helpful to say that here.

src/stan-users-guide/wiener_diffusion_model.qmd
```diff
@@ -0,0 +1,337 @@
...
+Depending on the experimental design, one would typically also provide the number of conditions and the condition associated with each trial as a vector. In a hierarchical setting, the **data block** would also specify the number of participants and the participant associated with each trial as a vector. It is also possible to hand over a precision value in the **data block**.
+
```
bob-carpenter, line 195: What is a condition?  Is that something that's going to be represented by a covariate with some kind of regression?

src/stan-users-guide/wiener_diffusion_model.qmd
```diff
@@ -0,0 +1,337 @@
...
+
```
bob-carpenter, line 196: Emphasis with italics, `* ... *` or `_ ... _`, not `**...**`

Having said that, names of blocks should not be bolded or even italicized here---they're just being used normally.  Same for stress on parameters block.

This is all picky little details to try to keep the docs consistent, so if you don't want to do this, because it's not like they're super consistent, feel free to just say "won't do".  We can always take passes later for consistency of things like headers.

src/stan-users-guide/wiener_diffusion_model.qmd
```diff
@@ -0,0 +1,337 @@
...
+  if (resp[i] == 1) { // upper boundary
+   target += wiener_full_lpdf(rt[i] | a, t0, w, v, sv, sw, st0);
+  } else { // lower boundary(mirror drift and starting point )
+   target += wiener_full_lpdf(rt[i] | a, t0, 1-w, -v, sv, sw, st0);
+  }
+ }
+}
+```
+
+### The Data Block
```
bob-carpenter: I've tried to use sentence case for headers throughout, so this would be "The data block" and so on.

src/stan-users-guide/wiener_diffusion_model.qmd
```diff
@@ -0,0 +1,337 @@
...
+
+
+
+### The Parameters Block
+
+The model arguments of the `wiener_lpdf()` function that are not fixed to a certain value are defned as parameters in the **parameters block**. In this block, it is also possible to insert restrictions on the parameters. Note that the MCMC algorithm iteratively searches for the next parameter set. If the suggested sample falls outside the internally defined parameter ranges, the program will throw an error, which causes the algorithm to restart the current iteration. Since this slows down the sampling process, it is advisable to include the parameter ranges in the defnition of the parameters in the **parameters block** to improve the sampling process (see table below for the parameter ranges). In addition, the parameter space is further constrained by the following conditions:
+
+1. The non-decision time $t_0$ has to be smaller or equal to the observed reaction time: $t0 \leq y$.
```
bob-carpenter: This should go into the code as a constraint.  I'm not 100% sure how to write it because of the flipping from positive to negative.

Maybe something like

```stan
real<lower=0, upper=min_rt> t0;
```

where `min_rt` is defined as `min(rt)` in transformed data (so it's only performed once).  Because `rt` is fixed data, we don't need to use it in truncation of the distribution on `t0`.

There are also many other constraints on parameters that aren't mentioned here.

src/stan-users-guide/wiener_diffusion_model.qmd
```diff
@@ -0,0 +1,337 @@
...
+
+ // diffusion model
+ for (i in 1:N) {
+  if (resp[i] == 1) { // upper boundary
+   target += wiener_full_lpdf(rt[i] | a, t0, w, v, sv, sw, st0);
+  } else { // lower boundary(mirror drift and starting point )
+   target += wiener_full_lpdf(rt[i] | a, t0, 1-w, -v, sv, sw, st0);
+  }
+ }
+}
```
bob-carpenter, line 155: "non-decision" doesn't get extra spaces around hyphens.

src/stan-users-guide/wiener_diffusion_model.qmd
```diff
@@ -0,0 +1,337 @@
...
+
+
+
+### The Parameters Block
+
+The model arguments of the `wiener_lpdf()` function that are not fixed to a certain value are defned as parameters in the **parameters block**. In this block, it is also possible to insert restrictions on the parameters. Note that the MCMC algorithm iteratively searches for the next parameter set. If the suggested sample falls outside the internally defined parameter ranges, the program will throw an error, which causes the algorithm to restart the current iteration. Since this slows down the sampling process, it is advisable to include the parameter ranges in the defnition of the parameters in the **parameters block** to improve the sampling process (see table below for the parameter ranges). In addition, the parameter space is further constrained by the following conditions:
+
+1. The non-decision time $t_0$ has to be smaller or equal to the observed reaction time: $t0 \leq y$.
+2. The varying relative starting point $w$ has to be in the interval (0,1) and thus,
+
```
bob-carpenter, line 217: These should also be in the constraints to avoid rejections, i.e.,

```stan
real<lower=0, upper=fmin(2 * w, 2 * (1 - w))> s_w;
```

And it should go in the table.

src/stan-users-guide/wiener_diffusion_model.qmd
```diff
@@ -0,0 +1,337 @@
...
+|$a$         | (0, $\infty$)|   | $y$         |(0, $\infty$)|
+|$v$         | (-$\infty$, $\infty$)| | $s_v$ |[0, $\infty$)|
+|$w$         | (0,1)        |   | $s_w$     | [0,1)|
+|$t_0$       |[0,$\infty$)  |   | $s_{t0}$  |[0,$\infty$)|
+
+
+
+### The Model Block
+
+In the **model block**, the priors are defined and the likelihood is called for the upper and the lower response boundary. Different kinds of priors can be specifed here. When no prior is specifed for a parameter, Stan uses default priors.
```
bob-carpenter: Again, no bold for the block names---they're just regular usages.

src/stan-users-guide/wiener_diffusion_model.qmd
```diff
@@ -0,0 +1,337 @@
...
+
```
bob-carpenter, line 241: Stan always starts with flat priors over the legal values of parameters.  Those may be improper if there's unbounded support.  Other priors just get multiplied in (or added on the log scale), which given that flat priors don't contribute anything, is like just specifying the stated priors.

src/stan-users-guide/wiener_diffusion_model.qmd
```diff
@@ -0,0 +1,337 @@
...
+Generally, mildly informative priors might help to get the full benefit of a Bayesian analysis.
+
+In the second part of the **model block**, the likelihood function is applied to all responses. Drift rate $v$ and relative starting point $w$ have to be mirrored for responses at the lower boundary.
+
```
bob-carpenter: Drift rate -> The drift rate

src/stan-users-guide/wiener_diffusion_model.qmd
```diff
@@ -0,0 +1,337 @@
...
+For more details, see @Henrich2024.
```
bob-carpenter: Is it possible to say what the details provide?  Is it prior recommendations, or what?

src/stan-users-guide/wiener_diffusion_model.qmd
```diff
@@ -0,0 +1,337 @@
...
+
+Depending on the implementation of the response window, two different types of data arise: _truncated_ data or _censored_ data. Since the effects of truncation or censoring on summary statistics such as mean, median, standard deviation, and skewness is regularly too large to ignore [@Ulrich1994], data analysts are well advised to account for these effects.
+
+As described in the [Truncated or Censored Data](./truncation-censoring.qmd) chapter, the cumulative distribution function ($\text{CDF}$) and its complement ($\text{CCDF}$) are needed to model truncated and censored data.
+
+As explained above, the $\text{PDF}$ is defined _defectively_, meaning that only the sum of the $\text{PDF}$s for both response alternatives integrates to 1. For the same reason, the $\text{CDF}$ and $\text{CCDF}$ are also implemented defectively. Analogously, only the sum of the $\text{CDF}$s and $\text{CCDF}$s for both response alternatives converges to 1.
+
+In the case of the diffusion model, the $\text{CDF}$ converges to the probability $P$ to hit the corresponding response boundary: (for simplicity, we omit the inter-trial variabilities in the following)
+
```
bob-carpenter, line 283: converges to -> asymptotes above at 

Also, the symbol $P$ isn't doing anything here.

Going forward, I'd recommend using just about anything other than $P$ as we already have $p$ for densities and mass functions and $\Pr$ for event probabilities.

src/stan-users-guide/wiener_diffusion_model.qmd
```diff
@@ -0,0 +1,337 @@
...
+Then, the density of truncated data for both response boundaries 0 and 1, here denoted as $\text{resp}\in\{0,1\}$, can be formulated as follows:
+
+$$
+\begin{aligned}
+&\text{PDF}_{\text{resp}}(x \mid L<X\leq U, a, w, v) = \\ &\frac{\text{PDF}_{\text{resp}}(x \mid a, w, v)\cdot \mathbb{I}_{\{L<x\leq U\}}}
+{\bigl(\text{CDF}_0(U \mid a, w, v)+\text{CDF}_1(U \mid a, w, v)\bigr) -
+\bigl(\text{CDF}_0(L\mid a, w, v)+\text{CDF}_1(L\mid a, w, v)\bigr)}
+\end{aligned}
+$$
+
```
bob-carpenter, line 315: I think this is too much to change, but this is not how we've written pdfs and cdfs in the rest of the documentation.  Usually a pdf is just $p(x)$ and $p_X(x)$ if you want to indicate the random variable $X$ for which it's a pdf.  Cdfs are then written as $F_X$, following the usual convention in stats.  If they're conditioned, that'd be $F_{X | Y}$ and so on.  Anyway, no need to change this unless you're really motivated.

src/stan-users-guide/wiener_diffusion_model.qmd
```diff
@@ -0,0 +1,337 @@
...
+   wiener_lcdf_unnorm(left_bound | a, t0, 1-w, -v, sv, sw, st))
+ ); // parenthesis log_diff_exp
+ // compute log - likelihood
+ for (i in 1:N) {
+  if (resp[i] == 1) { // response -1 boundary
+   target += wiener_lpdf (rt[i] | a, t0, w, v, sv, sw, st);
+  } else { // response -0 boundary ( mirror v and w)
+   target += wiener_lpdf (rt[i] | a, t0, 1-w, -v, sv, sw, st);
+  }
+   target += -denom ;
```
bob-carpenter: This can be moved out of the loop and replaced with

```stan
target += -N * log_denom;
```

Note that it's a log denominator, so I changed notation, not the denominator you wrote down.

src/stan-users-guide/wiener_diffusion_model.qmd
```diff
@@ -0,0 +1,337 @@
...
+  log_sum_exp(
+   wiener_lcdf_unnorm(right_bound | a, t0, w, v, sv, sw, st),
+   wiener_lcdf_unnorm(right_bound | a, t0, 1-w, -v, sv, sw, st)),
+  log_sum_exp(
+   wiener_lcdf_unnorm(left_bound | a, t0, w, v, sv, sw, st),
+   wiener_lcdf_unnorm(left_bound | a, t0, 1-w, -v, sv, sw, st))
+ ); // parenthesis log_diff_exp
+ // compute log - likelihood
+ for (i in 1:N) {
+  if (resp[i] == 1) { // response -1 boundary
```
bob-carpenter: indentation should be two spaces, not one

src/stan-users-guide/wiener_diffusion_model.qmd
```diff
@@ -0,0 +1,337 @@
...
+   target += wiener_lpdf (rt[i] | a, t0, w, v, sv, sw, st);
+  } else { // response -0 boundary ( mirror v and w)
+   target += wiener_lpdf (rt[i] | a, t0, 1-w, -v, sv, sw, st);
+  }
+   target += -denom ;
+ } // end for
+}
+```
+
+How to call a truncated model within the parallelization routine of `reduce_sum`or with truncation to only on side, see @Henrich2026.
```
bob-carpenter: How to call -> For details of how to call

It's not grammatical in English as written.

src/stan-users-guide/wiener_diffusion_model.qmd
```diff
@@ -0,0 +1,337 @@
...
+
+For the censored model, we distinguish two cases: In the first case, the responses of the censored trials are known, but the reaction times are not known. In the second case, neither the responses nor the reaction times of the censored trials are known. Note that the second case differs from a truncated model in the fact that the number of censored trials is still known. Consider first the case where the response is known even for censored data.
+
+To model such data in Stan, the left and right reaction time bounds, `left_bound` and `right_bound`, respectively, are handed over in the **data block**, as well as a vector `censored` that tracks whether a trial is censored (= 1) or not (= 0), and counts of trials censored at the left reaction time bound and counts of trials censored at the right reaction time bound for each response in {0,1}. There are four such count variables: `N_cens_left_0`, `N_cens_left_1`, `N_cens_right_0`, `N_cens_right_1`:
+
+
+```stan
+model { // ... // definition of priors for all model parameters
```
bob-carpenter: two spaces indentation.  And you can break over lines as follows (unlike in R, which doesn't understand line breaks, Stan can follow the standard mathematical convention of moving the operator down to the next line),

```stan
  target += N_cens_left_0
         * wiener_lcdf_unnorm(left_bound | a, t0, 1-w, -v, sv, sw, st0);
```

src/stan-users-guide/wiener_diffusion_model.qmd
```diff
@@ -0,0 +1,337 @@
...
+ target += N_cens_right_0 * wiener_lccdf_unnorm(right_bound | a, t0, 1-w, -v, sv, sw, st0);
+
+ // summands for response = 1
+ target += N_cens_left_1 * wiener_lcdf_unnorm(left_bound | a, t0, w, v, sv, sw, st0);
+
+ target += N_cens_right_1 * wiener_lccdf_unnorm(right_bound | a, t0, w, v, sv, sw, st0);
+}
+```
+
+When data are censored at only one side, omit the lines for the other side in the code.
```
bob-carpenter: I think this means when the censoring counts are zero, the relevant target increment statements can be dropped to save computation.

src/stan-users-guide/wiener_diffusion_model.qmd
```diff
@@ -0,0 +1,337 @@
...
+
+When data consist of many conditions, it is sometimes more convenient to loop over all trials instead of using count variables as described above, using the following notation and code. A vector containing the information whether a trial is censored or not, here `censored`, needs to be handed over in the **data block**. This vector splits the data into three bins: all trials $i$ with `censored[i]=0` are censored below the left reaction time bound, all trials $i$ with `censored[i]=1` fall between the reaction time bounds, and all trials $i$ with `censored[i]=2` are censored above the right reaction time bound. For non-censored trials, the log-PDF is computed, for left censored trials, the log-CDF is computed, and for right censored trials, the log-CCDF is computed:
```
bob-carpenter: Does "many conditions" mean anything other than different truncation/censoring boundaries?   If so, swaying what is meant by a condition would help, and if not, just saying "different boundaries" would help.

src/stan-users-guide/wiener_diffusion_model.qmd
```diff
@@ -0,0 +1,337 @@
...
+
+
+```stan
+model { // ... // definition of priors for all model parameters
+ for (i in 1:N) { // right censored at right_bound
+  if (resp [i] == 1) { // upper response boundary
+   if (censored[i] == 1) {
```
bob-carpenter: I'd find this easier to scan if it went 0, 1, 2 in order.

src/stan-users-guide/wiener_diffusion_model.qmd
```diff
@@ -0,0 +1,337 @@
...
+    target += wiener_lcdf_unnorm(left_bound | a, t0, 1-w, -v, sv, sw, st0);
+   } else if (censored[i] == 2) {
+    target += wiener_lccdf_unnorm(right_bound | a, t0, 1-w, -v, sv, sw, st0);
+   }
+  }
+ }
+}
+```
+
+When the data are censored to only one side, omit the case that is not needed.
```
bob-carpenter: censored to -> censored on

Sorry preposition choice makes zero sense in languages.

src/stan-users-guide/wiener_diffusion_model.qmd
```diff
@@ -0,0 +1,337 @@
...
+
+Note that this block can be inserted in the defnition of the parallelization function, `partial_sum_wiener()`, as defined below.
+
+Censoring sometimes includes the response (i.e., it is known that the reaction time in a trial fell outside the response window, but which response was given is unknown). One method that has been used to model such data has involved inferring the numbers of missing responses of either kind from the observed relative frequencies of the two responses. This approach has the problem that quite specifc assumptions on the missing data have to be made (namely, that the proportions of the two kinds of responses are the same for responses within and outside the response window).
+
```
bob-carpenter: The first sentence is confusing.  If the data includes the response, isn't the response known?  

Also, I wouldn't say that "censoring sometimes includes the response"---the time is the variable being censored.  If the response isn't observed, it would be called "missing".

src/stan-users-guide/wiener_diffusion_model.qmd
```diff
@@ -0,0 +1,337 @@
...
+Here is a more principled approach that uses the cumulative distribution functions and their complements to provide the likelihood of censored data. As before, let $L$ be the left reaction time bound, and $U$ the right reaction time bound, and consider decision times without inter-trial variabilities for the sake of simplicity. It follows that the likelihood $p_l$ of observing a left-censored data point
+is given by
```
bob-carpenter: This isn't quite the right usage of "likelihood".  See:

https://statmodeling.stat.columbia.edu/2026/03/20/a-data-model-is-not-just-a-likelihood/

If we have a data-generating distribution $p(y \mid \theta)$, then if we fix $\theta$, we get a density for $y$.  If we fix $y = y^\text{obs}$, we get the likelihood function $\mathcal{L}(\theta) = p(y^\text{obs} \mid \theta)$.  

Specifically, this means we can't say "likelihood of data" as the likelihood function, in order to exist, has already fixed the data.

This is confused in the literature all over the place, but *BDA* makes it clear, as does Aki's post.

src/stan-users-guide/wiener_diffusion_model.qmd
```diff
@@ -0,0 +1,337 @@
...
+
+$$
+\begin{aligned}
+p_r(a,w,v) = \text{CCDF}_0(U\mid a,w,v) + \text{CCDF}_1(U\mid a,w,v).
+\end{aligned}
+$$
+
+
+See the following code for an example of Stan code implementing this second case of censoring. This model call deals with the problem of unknown responses by computing the probability of choosing the response-1- or response-0 boundary outside the response window. Here, the CDF and/or the CCDF are required, depending upon whether there is only left-censoring, right-censoring, or censoring both to the left and to the right. The following code shows the **functions block** for a model that is right-censored using the function `partial_sum_wiener()` for parallel computations. Combine this block with the **model block** in the example above:
+
```
bob-carpenter, line 539: he block doesn't actually show any parallel computations, so either get rid of this or say a bit more about what would be required.

src/stan-users-guide/wiener_diffusion_model.qmd
```diff
@@ -0,0 +1,337 @@
...
+```stan
+functions { // parallelization function
```
bob-carpenter: It's not clear what "parallelization function" refers to here---whatever it is, it shouldn't be a comment on the functions block.  

I take it you mean that the `partial_sum_wiener` function could be used in `reduce_sum`?  If so, it would help to say so somewhere.

src/stan-users-guide/wiener_diffusion_model.qmd
```diff
@@ -0,0 +1,337 @@
...
+ real partial_sum_wiener(array[] real rt_slice, int start,
+ int end, real a, real t0, real w, real v, real sv, real sw,
```
bob-carpenter: indent the second and third line of arguments under the first arguments.

two spaces indent everywhere

src/stan-users-guide/wiener_diffusion_model.qmd
```diff
@@ -0,0 +1,337 @@
...
+   } else { // censored data
+    ans += log_sum_exp (
+     wiener_lccdf_unnorm(right_bound | a, t0, w, v, sv, sw, st),
+     wiener_lccdf_unnorm(right_bound | a, t0, 1-w, -v, sv, sw, st);
+   }
+  } // end for
+  return ans;
+ } // end partial_sum_fullddm
+} // end functions
+```
```
bob-carpenter, line 571: Like the earlier functions, a transformed data block that sorted the three types of data would allow vectorization and allow the censoring parts to be replaced. with a product given that the right bound doesn't change.

bob-carpenter (changes requested):
Hi, Franzi:

I'm really sorry this is so many comments.  Most of it's to try to get consistency with the rest of our documentation's formatting.

The most important comment is about the upper bounds for parameters that are missing from the examples.

If you'd rather not make the fixes yourself, I think I can go through and make almost all of them other than places where I had genuine questions.

src/bibtex/all.bib
Franzi2114: The image is from my first paper. How do I cite this correctly in the qmd?

Franzi2114: The pdf is of much better quality. The jpg quickly becomes pixled. Therefore, I would prefer to only include the pdf-version and delete the jpg version. 

If that is not possible and you prefer the jpg, I can delete the pdf version. What do you think?

src/stan-users-guide/wiener_diffusion_model.qmd
Franzi2114: I added this to the bib.all. How shall I cite the Chapter?

I now added the book citation behind Ratcliff:
"... non-decision time, $s_{t0}$  [@Ratcliff1998, @nicenboim2025introduction]."

src/stan-users-guide/wiener_diffusion_model.qmd
Franzi2114: Yes, the model still works if variances are set to zero. All these cases are handled in the algorithm.

src/stan-users-guide/wiener_diffusion_model.qmd
Franzi2114: No, this is the general explanation for the coding. In both cases we have "upper bound " and "lower bound". In accuracy coding, the bounds correspond to correct/error, and in response coding, the bounds correspond to Response 1/Response 2.

I added another sentence:
"This means, depending on the coding you 
choose, the bounds mentioned in the second variable above differ and the 
_response_ variable will have a different form."

src/stan-users-guide/wiener_diffusion_model.qmd
Franzi2114: No. A condition will be represented by an own model. Data will be split by condition and for each condition an own model will be estimated. This means, the parameters that are going to be estimated are not defined only as parameters with one entry, but instead as vectors or even matrices.

A condition is for example the following: One group with manipulation, and one group without manipulation.

src/stan-users-guide/wiener_diffusion_model.qmd
Franzi2114: Yes, this is not the  conventional way of the implementation of this density that you propose.
Usually, there is one variable that contains the reaction times in seconds and second variable that contains the responses coded in 0 and 1.
Depending on the response, the density for the upper bound or the lower bound will be computed. The only difference between both bounds is -v instead of v and 1-w instead of w.

src/stan-users-guide/wiener_diffusion_model.qmd
Franzi2114: changed

src/stan-users-guide/wiener_diffusion_model.qmd
Franzi2114: Added here.
However, this is also built in the code for the case that users forget this.

src/stan-users-guide/wiener_diffusion_model.qmd
Franzi2114: Details regarding the application of the diffusion model in Stan.

src/stan-users-guide/wiener_diffusion_model.qmd
Franzi2114: I called it PROB.  Open for alternatives.

src/stan-users-guide/wiener_diffusion_model.qmd
Franzi2114: I changed this. Is there also another name for the CCDFs?

src/stan-users-guide/wiener_diffusion_model.qmd
Franzi2114: "Variability" is the technical term that is also used in the diffusion model literature. Therefore, I would like to stick to this. It is directly linked to the mathematical terms of variance or standard deviation. 

I added another explanatory sentence on that.

src/stan-users-guide/wiener_diffusion_model.qmd
Franzi2114: The term "first-passage time" refers to stochastic processes where the random walk crosses a threshold for the first time. I will add another sentence on that.

Yes, the reaction time shall be positive, continuous and in seconds. For the diffusion model, we also need a vector for the responses that codes which of the decisions was taken. Then, the user implements in the model by inserting `v` and `w`, or  `-v` and `1-w` which of the two densities to compute. I also added a sentence on that.

src/stan-users-guide/wiener_diffusion_model.qmd
Franzi2114: What do you mean here? Which things do not align?

src/stan-users-guide/wiener_diffusion_model.qmd
Franzi2114: I think we prepared everything for the vectorization but it was not yet possible to vectorize over so many parameters. Therefore, I propose to use the for-loops everywhere in the function calls.

src/stan-users-guide/wiener_diffusion_model.qmd
Franzi2114: No, this means when there is only one side of the reaction time continuum censored. In experiments, there might be reaction time windows: Only responses between 200 and 800 ms are accepted. Here, both sides are censored. Sometimes, there is only an upper bound: Only responses below 800 ms are accepted. Then, data is only censored to one side.
I added a sentence on that.

src/stan-users-guide/wiener_diffusion_model.qmd
Franzi2114: No, here I refer to experimental conditions. I added some explanatory sentences.

src/stan-users-guide/wiener_diffusion_model.qmd
Franzi2114: What I meant is the following `parallelize the execution of a single Stan chain across multiple cores`. I changed the sentence accordingly.

src/stan-users-guide/wiener_diffusion_model.qmd
Franzi2114: As described above, I think that vectorization is prepared but not yet possible.

src/stan-users-guide/wiener_diffusion_model.qmd
Franzi2114: Which formulation would you prefer here?

Franzi2114:
Dear @bob-carpenter, I worked through all your comments and made several changes!

Could you have another look at it? I also had some questions regarding your posts. I wrote them below your comments.

I am afraid that I messed something up with the all.bib file. Can you fix this?

bob-carpenter:
There are also conflicts with character usage of unicode ~ when you try to build.  Do all the docs build locally for you with the broken all.bib file and the stray unicode characters?

It looks somehow like you dumped in an entirely conflicting bibtex file somehow.  To fix, you just need to get rid of all the notes about conflicts that were autogenerated in `all.bib` and leave just what should be there, then push.  I can do that locally.  

I'll just go ahead and make the language changes and indicate what I did above.   The main issue is that likelihoods are functions of parameters, not data, by construction.

src/stan-users-guide/wiener_diffusion_model.qmd
bob-carpenter: I just fixed the language to talk about the contribution to the likelihood function or to the data generating distribution, whichever made sense in context.  A data point can't technically be said to have a "liklelihood" because the likelihood is a function of parameters.  It's very confusing, because people say this kind of thing all the time.

src/bibtex/all.bib
bob-carpenter: I added this:  

>  Image from @Henrich2024, distributed under the Creative Commons Attribution 4.0 International License.

bob-carpenter: We need the jpg version for the web site version.  You can always generate a high-resolution jpg image from the pdf if you'd like.  But probably not bigger than 1MB or the page will get laggy.  To switch based on document type, I added this:

```qmd
format:
  html:
    default-image-extension: jpg
  pdf:
    default-image-extension: pdf
```

src/stan-users-guide/wiener_diffusion_model.qmd
bob-carpenter: That's good.

src/stan-users-guide/wiener_diffusion_model.qmd
bob-carpenter: If my explanation isn't clear, I'm not going to try to update the doc.  It's just that we have to be careful about saying what the data generating distribution is when the data is multivariate and heterogeneous (like boundary plus crossing time as it is here).

bob-carpenter:
There are non-ASCII versions of `~` on several lines.  That won't render in LaTeX.  I just switched it back to ASCII.

```bash
stan-users-guide (feature/docs_for_wiener_lpdf_and_lcdf_unnorm)$ grep -rn $'\xe2\x88\xbc' *.qmd

wiener_diffusion_model.qmd:127:  a ∼ normal(1, 1);
wiener_diffusion_model.qmd:128:  w ∼ normal(0.5, 0.1);
wiener_diffusion_model.qmd:129:  v ∼ normal(2, 3);
wiener_diffusion_model.qmd:130:  t0 ∼ normal(0.435, 0.12);
wiener_diffusion_model.qmd:132:  sv ∼ normal(1, 3);
wiener_diffusion_model.qmd:133:  st0 ∼ normal(0.183, 0.09);
wiener_diffusion_model.qmd:134:  sw ∼ beta(1, 3);
wiener_diffusion_model.qmd:384:        y[i] ∼ wiener(a, t0, w, v, sv, sw, st0);
wiener_diffusion_model.qmd:388:        y[i] ∼ wiener(a, t0, one_minus_w, neg_v, sv, sw, st0);
```

I don't actually see any conflicts in the `all.bib` file on the branch.

I went through and cut lines down to 80 characters.  I also spaced out operators, etc. in code and tightened spaces to conform to our coding guidelines (appendix of user's guide). I also removed the "end" block and other comments on lines with braces and moved them down so the structure of the code is clearer.

Italic words shouldn't be in formulas like $lik_i$---it needs to be $\textit{lik}_i$ so that "lik" is typeset like text rather than like math, the latter of which tends to insert odd spacing based on its guess as to what's a function and argument.

bob-carpenter:
I just went ahead and converted the figure to a .png and made that the default format for the web page.

I just realized in doing a diff on the command line that the conflict with the .bib file is that it got Windows-style newlines.  We strictly use Unix/Linux/Mac newlines.  Windows should *not* have tried to innovate on the newline character!  I switched it back using `dos2unix`.

It all builds now and I checked the output in html and pdf and the images look right.

Thanks for bearing with the long process here.  It only took me an hour or so to put in the final patches.

bob-carpenter (approved):
Looks good. Thanks for all the updates.

PR #939, opened 2026-03-26
Update tuple doc to allow for length one tuples

WardBrian:
#### Submission Checklist

- [x] Builds locally

#### Summary

Doc for https://github.com/stan-dev/stanc3/pull/1609

#### Copyright and Licensing

Simons Foundation

### DIFF
--- a/src/reference-manual/expressions.qmd
+++ b/src/reference-manual/expressions.qmd
@@ -401,8 +401,9 @@ The expression `42` is of type `int` and the expression `[1, 2.9,
 -1.3]')` thus has a sized type of `tuple(int, vector[3])` and an
 unsized type (e.g., for a function argument) of `tuple(int, vector)`.
 
-Stan does *not* support the Python notation with trailing commas, such
-as `(1, 2, 3, )` for a 3-tuple.
+While they are rarely useful, a tuple of one element can be created using
+the same style as languages like Python, with a trailing comma, e.g.
+`(3.14,)`. For longer tuples, Stan does *not* support trailing commas.
 
 ### Restrictions on values {-}
 
@@ -446,12 +447,14 @@ zero-element array.
 array[0] int a;   // a is fully defined as zero element array
 ```
 
-#### No zero-tuples or one-tuples {-}
+#### No zero-tuples {-}
 
-There is no way to declare or construct a zero-tuple or one-tuple in
-Stan.  Tuples must be at least two elements long.  The expression `()`
-does not pick out a zero-tuple---it is ill formed.  Similarly, the
-expression `(1)` is of type `int` rather than a tuple.
+There is no way to declare or construct a zero-tuple in Stan.  Tuples must be at
+least one element long. The expression `()` does not pick out a zero-tuple---it
+is ill formed.
+
+One-tuples need a trailing comma, like `(1,)`. The expression `(1)`, without the
+comma, is of type `int` rather than a tuple.
 
 ## Parentheses for grouping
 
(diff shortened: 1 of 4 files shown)


src/reference-manual/expressions.qmd
```diff
@@ -401,8 +401,9 @@ The expression `42` is of type `int` and the expression `[1, 2.9,
 -1.3]')` thus has a sized type of `tuple(int, vector[3])` and an
 unsized type (e.g., for a function argument) of `tuple(int, vector)`.
 
-Stan does *not* support the Python notation with trailing commas, such
-as `(1, 2, 3, )` for a 3-tuple.
+While they are rarely useful, a tuple of one element can be created using
+the same style as languages like Python, with a trailing comma, e.g.
```
bob-carpenter: "e.g." is always followed by a comma

src/reference-manual/expressions.qmd
```diff
@@ -401,8 +401,9 @@ The expression `42` is of type `int` and the expression `[1, 2.9,
 -1.3]')` thus has a sized type of `tuple(int, vector[3])` and an
 unsized type (e.g., for a function argument) of `tuple(int, vector)`.
 
-Stan does *not* support the Python notation with trailing commas, such
-as `(1, 2, 3, )` for a 3-tuple.
+While they are rarely useful, a tuple of one element can be created using
```
bob-carpenter: I'd remove the editorializing about utility here, but I'll leave it up to you.

src/reference-manual/expressions.qmd
```diff
@@ -446,12 +447,14 @@ zero-element array.
...
 
-#### No zero-tuples or one-tuples {-}
+#### No zero-tuples {-}
 
-There is no way to declare or construct a zero-tuple or one-tuple in
-Stan.  Tuples must be at least two elements long.  The expression `()`
-does not pick out a zero-tuple---it is ill formed.  Similarly, the
-expression `(1)` is of type `int` rather than a tuple.
+There is no way to declare or construct a zero-tuple in Stan.  Tuples must be at
+least one element long. The expression `()` does not pick out a zero-tuple---it
```
bob-carpenter, line 453: [QUESTION]
Is there a reason to disallow `(,)`?  I'd have thought it'd actually be easier syntactically?

src/reference-manual/types.qmd
```diff
@@ -1436,11 +1436,17 @@ may be declared as
...
 
 ```stan
 tuple() nil;  // ILLEGAL
+```
+
+Tuples of length one must use a trailing comma, to align with the expression
+syntax.
+
+```stan
+tuple(int,) one; // CORRECT
```
bob-carpenter: How about `m` instead of `one` to match h the following?  I think using "one" can be confusing as a variable name.

bob-carpenter (changes requested):

src/reference-manual/expressions.qmd
WardBrian: It's actually slightly more complicated. With the current rules, we can parse tuples as 

```
tuple_expr := "(" expression "," (expression ("," expression)*)? ")"
```

If we wanted to allow `(,)` (ignoring that a statically-zero-sized type is always useless), we'd need to do something like the above plus `| "(" "," ")"`

Also, besides, `()` would actually unambiguously parse in Stan, so if we wanted the zero tuple we'd probably just parse it as that, which is what e.g. Python does. The special-case trailing-comma really only is obligatory for the length 1 tuple

bob-carpenter: Thanks for the explanation.  I didn't realize Python took `()` to be the empty tuple---it makes sense that would have a unique parse.  It just breaks my sense of consistency.

bob-carpenter (approved):
Thanks!

PR #974, opened 2026-09-02
Document poisson-binomial distribution

WardBrian:
#### Submission Checklist

- [x] Builds locally

#### Summary

Documentation for https://github.com/stan-dev/stanc3/pull/1690

#### Copyright and Licensing

Simons Foundation

### DIFF
--- a/src/functions-reference/bounded_discrete_distributions.qmd
+++ b/src/functions-reference/bounded_discrete_distributions.qmd
@@ -353,6 +353,92 @@ For a description of argument and return types, see section
 [vectorized PRNG functions](conventions_for_probability_functions.qmd#prng-vectorization).
 {{< since 2.18 >}}
 
+## Poisson-binomial distribution {#poisson-binomial}
+
+### Probability mass function
+
+Suppose $N \in \mathbb{N}$, $y \in \{0,\ldots,N\}$, and $\theta \in [0,1]^N$. Let
+$F_y$ denote the set of all length-$y$ subsets of $\{0,\ldots,N\}$.
+Then, \begin{equation*} \text{PoissonBinomial}(y~|~\theta) = \sum\limits_{A\in F_y} \prod\limits_{i\in A} \theta_i \prod\limits_{j\in A^c} (1-\theta_j). \end{equation*}
+
+
+### Distribution statement
+
+
+`y ~ ` **`poisson_binomial`**`(theta)`
+
+Increment target log probability density with `poisson_binomial_lupmf(y | theta)`.
+{{< since 2.40 >}}
+<!-- real; poisson_binomial ~; -->
+\index{{\tt \bfseries poisson\_binomial }!distribution statement|hyperpage}
+
+### Stan functions
+
+<!-- real; poisson_binomial_lpmf; (ints y | vectors theta); -->
+\index{{\tt \bfseries poisson\_binomial\_lpmf }!{\tt (ints y \textbar vectors theta): real}|hyperpage}
+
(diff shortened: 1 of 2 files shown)


src/functions-reference/bounded_discrete_distributions.qmd
```diff
@@ -353,6 +353,92 @@ For a description of argument and return types, see section
 [vectorized PRNG functions](conventions_for_probability_functions.qmd#prng-vectorization).
 {{< since 2.18 >}}
 
+## Poisson-binomial distribution {#poisson-binomial}
+
+### Probability mass function
+
+Suppose $N \in \mathbb{N}$, $y \in \{0,\ldots,N\}$, and $\theta \in [0,1]^N$. Let
+$F_y$ denote the set of all length-$y$ subsets of $\{0,\ldots,N\}$.
```
bob-carpenter: length -> size (subsets don't have a length).

y -> n (we usually use y for continuous variables and n for discrete ones---or k or j or some middle letter like that)

I think another notation than F_y would be good because F_X is the conventional notation for a cdf for the random variable X.

src/functions-reference/bounded_discrete_distributions.qmd
```diff
@@ -353,6 +353,92 @@ For a description of argument and return types, see section
...
+Then, \begin{equation*} \text{PoissonBinomial}(y~|~\theta) = \sum\limits_{A\in F_y} \prod\limits_{i\in A} \theta_i \prod\limits_{j\in A^c} (1-\theta_j). \end{equation*}
```
bob-carpenter: I don't actually know what \limits does, but you can just get rid of it here.  

Also, I've been using \mid rather than ~|~, though what you have is better here because it should prevent breaks just before or after the bar.

src/functions-reference/bounded_discrete_distributions.qmd
bob-carpenter: What is the complexity implication here given that F_y is size (N choose y)?

src/functions-reference/bounded_discrete_distributions.qmd
```diff
@@ -353,6 +353,92 @@ For a description of argument and return types, see section
 [vectorized PRNG functions](conventions_for_probability_functions.qmd#prng-vectorization).
 {{< since 2.18 >}}
 
+## Poisson-binomial distribution {#poisson-binomial}
+
+### Probability mass function
```
bob-carpenter, line 362: We might want to add that this is also called the "generalized binomial distribution" and mention its motivation as the distribution of successes when each of N trials has a different probability given by theta_n.  

It's also worth noting here that the computation isn't as bad as the (N choose y) might have you believe because there's a good dynamic programming algorithm for this (I coded it on the forums in the past).

bob-carpenter (changes requested):
Thanks, this looks great.  I hadn't realized this was added.  Just a couple clarifications requested in the doc and the notation.

src/functions-reference/bounded_discrete_distributions.qmd
WardBrian: The dynamic programming algorithm is the one used [in `math`](https://github.com/stan-dev/math/blob/develop/stan/math/prim/fun/poisson_binomial_log_probs.hpp)

WardBrian:
>  I hadn't realized this was added

It had a long journey. Original math implementation was in 2020, it had some issues until 2025, and then slipped through the cracks of actually being exposed in the language until I did a sweep through old issues this week

bob-carpenter (approved):
Thanks for adding the explanation and simplifying the LaTeX.

